PR #15, opened 2022-09-23
Make our container image "distroless"

boxofrad:
Switches our release base image to harden our security posture.

See here: https://github.com/GoogleContainerTools/distroless

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,7 +6,7 @@
 
 # envoy-binary pulls in the latest Envoy binary, as Envoy don't publish
 # prebuilt binaries in any other form.
-FROM envoyproxy/envoy-distroless:v1.23.1 as envoy-binary
+FROM envoyproxy/envoy-distroless:v1.24.0 as envoy-binary
 
 # go-discover builds the discover binary (which we don't currently publish
 # either).


david-yu:
Thanks we need to wait on testing with Consul K8s before merging this in, this can be done after beta.

cmd/consul-dataplane/main.go
```diff
@@ -245,3 +248,19 @@ func main() {
...
+// all flag parsing here).
+func readServiceIDFromFile() {
+	if serviceID == "" && serviceIDPath != "" {
+		id, err := os.ReadFile(serviceIDPath)
+		if err != nil {
+			log.Fatalf("failed to read given -proxy-service-id-path: %v", err)
+		}
+		serviceID = string(id)
+	}
+}
```
boxofrad, line 266: Hey @ishustava @thisisnotashwin 👋🏻 

Do you have an explanation of why consul-k8s needs this, rather than putting the service ID directly in the CLI flags, that I could put here?

cmd/consul-dataplane/main.go
```diff
@@ -245,3 +248,19 @@ func main() {
 		log.Fatal(err)
 	}
 }
+
+// readServiceIDFromFile reads the service ID from the file specified by the
+// -proxy-service-id-path flag.
+//
+// We do this here, rather than in the consuldp package's config handling,
+// because this option only really makes sense as a CLI flag (and we handle
+// all flag parsing here).
```
boxofrad, line 257: I don't _think_ it'd make sense to add this to the config file, what do you think?

cmd/consul-dataplane/main.go
thisisnotashwin: In K8s, we often have a different container responsible for getting the service ID to begin with. Information across containers can only be shared via files on a shared volume. It makes it significantly easier to ready what is going on if we are dealing with filepath. Putting the service ID directly to the CLI would require reading the contents of that file anyway.

riddhi89 (approved):
LGTM

david-yu:
Small suggestion, the distress envoy container is actually `envoyproxy/envoy-distroless:v1.24.0` the one you are currently copying for a multi-stage build is from the ubuntu based envoy image.

boxofrad:
Thanks, @david-yu! I think the Envoy binary itself is actually the same in the distroless and Ubuntu-based images (as it's built in a previous step).

That said, we could certainly shave some time off our builds by using the distroless base 👍🏻

david-yu (approved):
Thanks, I assume we can merge after the next Consul K8s beta and try it out?

PR #22, opened 2022-09-30
DNS transparent proxy

chapmanc:
This PR adds support for a transparent DNS proxy to consul for both udp and tcp. It is dependent on [this](https://github.com/hashicorp/consul/pull/14811) being merged first.

This PR is best viewed commit by commit: 
1. [d3c9609](https://github.com/hashicorp/consul-dataplane/pull/22/commits/d3c960962248c37f0db942084117d93f1a0a1837) pulls in the [change](https://github.com/hashicorp/consul/pull/14811) from consul. 
2. [ee58710](https://github.com/hashicorp/consul-dataplane/pull/22/commits/ee587107b7c68f3c4a220eda12eb7dfa026725e9) adds in the dns module for listening on both udp and tcp
3. [63462c9](https://github.com/hashicorp/consul-dataplane/pull/22/commits/63462c9a84cad8ef45f0ce37210067dba180e405) wires up the dns proxy to the consul dataplane instance
4. [a488b44](https://github.com/hashicorp/consul-dataplane/pull/22/commits/a488b446ee5d66b990f50769c5db6801189c4e9e) Adds the ability to disable the server by default

The `consul-dns-bind-addr` and the `consul-dns-port` can configure the dns port that the proxy will listen on. If no consul-dns-port is set the server will be disabled. By default the server is disabled.

### DIFF
--- a/cmd/consul-dataplane/main.go
+++ b/cmd/consul-dataplane/main.go
@@ -116,7 +116,7 @@ func init() {
 	flag.BoolVar(&tlsInsecureSkipVerify, "tls-insecure-skip-verify", false, "Do not verify the server's certificate. Useful for testing, but not recommended for production.")
 
 	flag.StringVar(&consulDNSBindAddr, "consul-dns-bind-addr", "127.0.0.1", "The address that will be bound to the consul dns proxy.")
-	flag.IntVar(&consulDNSPort, "consul-dns-port", -1, "The port the consul dns proxy will listen on. By default 0 disables the dns proxy")
+	flag.IntVar(&consulDNSPort, "consul-dns-bind-port", -1, "The port the consul dns proxy will listen on. By default -1 disables the dns proxy")
 
 }
 
@@ -218,6 +218,7 @@ func main() {
 
 	err = consuldpInstance.Run(ctx)
 	if err != nil {
+		cancel()
 		log.Fatal(err)
 	}
 }
--- a/cmd/consul-dataplane/main.go
+++ b/cmd/consul-dataplane/main.go
@@ -56,6 +56,9 @@ var (
 
 	xdsBindAddr string
 	xdsBindPort int
+
+	consulDNSBindAddr string
+	consulDNSPort     int
 )
 
 func init() {
@@ -110,6 +113,10 @@ func init() {
 	flag.StringVar(&tlsKeyFile, "tls-key", "", "The path to a client private key file (only required if tls.grpc.verify_incoming is enabled on the server).")
 	flag.StringVar(&tlsServerName, "tls-server-name", "", "The hostname to expect in the server certificate's subject (required if -addresses isn't a DNS name).")
 	flag.BoolVar(&tlsInsecureSkipVerify, "tls-insecure-skip-verify", false, "Do not verify the server's certificate. Useful for testing, but not recommended for production.")
+
+	flag.StringVar(&consulDNSBindAddr, "consul-dns-bind-addr", "127.0.0.1", "The address that will be bound to the consul dns proxy.")
+	flag.IntVar(&consulDNSPort, "consul-dns-port", -1, "The port the consul dns proxy will listen on. By default 0 disables the dns proxy")
+
 }
 
 // validateFlags performs semantic validation of the flag values
@@ -187,6 +194,10 @@ func main() {
 		XDSServer: &consuldp.XDSServer{
 			BindAddress: xdsBindAddr,
 			BindPort:    xdsBindPort,
+		},
+		DNSServer: &consuldp.DNSServerConfig{
+			BindAddr: consulDNSBindAddr,
+			Port:     consulDNSPort,
 		},
 	}
 	consuldpInstance, err := consuldp.NewConsulDP(consuldpCfg)
--- a/pkg/consuldp/consul_dataplane.go
+++ b/pkg/consuldp/consul_dataplane.go
@@ -81,7 +81,7 @@ func validateConfig(cfg *Config) error {
 		return errors.New("envoy xDS bind address not specified")
 	case !strings.HasPrefix(cfg.XDSServer.BindAddress, "unix://") && !net.ParseIP(cfg.XDSServer.BindAddress).IsLoopback():
 		return errors.New("non-local xDS bind address not allowed")
-	case !strings.HasPrefix(cfg.DNSServer.BindAddr, "unix://") && !net.ParseIP(cfg.DNSServer.BindAddr).IsLoopback():
+	case cfg.DNSServer.Port != -1 && !net.ParseIP(cfg.DNSServer.BindAddr).IsLoopback():
 		return errors.New("non-local DNS proxy bind address not allowed")
 	}
 
--- a/pkg/consuldp/consul_dataplane.go
+++ b/pkg/consuldp/consul_dataplane.go
@@ -161,11 +161,6 @@ func (cdp *ConsulDataplane) Run(ctx context.Context) error {
 		return fmt.Errorf("failed to run proxy: %w", err)
 	}
 
-	if err = cdp.startDNSProxy(ctx); err != nil {
-		cdp.logger.Error("failed to start the dns proxy", "error", err)
-		return err
-	}
-
 	doneCh := make(chan error)
 	go func() {
 		select {
--- a/pkg/consuldp/consul_dataplane.go
+++ b/pkg/consuldp/consul_dataplane.go
@@ -81,6 +81,8 @@ func validateConfig(cfg *Config) error {
 		return errors.New("envoy xDS bind address not specified")
 	case !strings.HasPrefix(cfg.XDSServer.BindAddress, "unix://") && !net.ParseIP(cfg.XDSServer.BindAddress).IsLoopback():
 		return errors.New("non-local xDS bind address not allowed")
+	case !strings.HasPrefix(cfg.DNSServer.BindAddr, "unix://") && !net.ParseIP(cfg.DNSServer.BindAddr).IsLoopback():
+		return errors.New("non-local DNS proxy bind address not allowed")
 	}
 
 	creds := cfg.Consul.Credentials
@@ -154,13 +156,9 @@ func (cdp *ConsulDataplane) Run(ctx context.Context) error {
 		return fmt.Errorf("failed to run proxy: %w", err)
 	}
 
-	dnsServer, err := cdp.startDNSProxy(ctx)
-	if err != nil {
-		if err != dns.ErrServerDisabled {
-			cdp.logger.Error("failed to start the dns proxy", "error", err)
-			return err
-		}
-		cdp.logger.Info("dns proxy disabled: configure the Consul DNS port to enable")
+	if err = cdp.startDNSProxy(ctx); err != nil {
+		cdp.logger.Error("failed to start the dns proxy", "error", err)
+		return err
 	}
 
 	doneCh := make(chan error)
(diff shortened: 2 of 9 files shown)


pkg/dns/dns.go
```diff
@@ -0,0 +1,270 @@
...
+		err := d.connUDP.SetReadDeadline(time.Now().Add(time.Second * 10))
+		if err != nil {
+			logger.Error("failure to set read deadline on connection", "error", err)
+			return
+		}
+		bytesRead, addr, err := d.connUDP.ReadFrom(buf)
+		if err != nil {
+			if errors.Is(err, net.ErrClosed) {
+				logger.Info("connection closed")
+				break
```
loshz, line 136: I think this should be a hard `return`.

pkg/dns/dns.go
```diff
@@ -0,0 +1,270 @@
...
+			} else {
```
loshz, line 137: You don't need this `else`.

pkg/dns/dns.go
```diff
@@ -0,0 +1,270 @@
...
+}
+
+func (d *DNSServer) queryConsulAndRespondUDP(buf []byte, addr net.Addr) {
+	logger := d.logger.Named("udp")
+	req := &pbdns.QueryRequest{
+		Msg:      buf,
+		Protocol: pbdns.Protocol_PROTOCOL_UDP,
+	}
+
+	resp, err := d.client.Query(context.Background(), req)
```
loshz, line 158: Should the context have a timeout?

pkg/dns/dns.go
```diff
@@ -0,0 +1,270 @@
...
+				if nerr, ok := err.(net.Error); ok && nerr.Timeout() {
+					logger.Debug("timeout waiting for read", "error", err)
+				} else {
+					logger.Warn("error reading from conn", "error", err)
+				}
+				continue
+			}
+		}
+		// Parallelize responses
+		go d.queryConsulAndRespondUDP(buf[0:bytesRead], addr)
```
loshz, line 147: How many responses do we predict we'll need to send? Do we need to limit the concurrency here with a worker group or similar?

pkg/dns/dns.go
```diff
@@ -0,0 +1,270 @@
...
+
+		// This is a guard and shouldn't happen but if the response is > 65535
+		// then we will just close the connection.
+		// Source: RFC1035 4.2.2.
+		if len(resp.Msg) > math.MaxUint16 {
+			logger.Error("consul response too large for DNS spec", "error", err)
+			return
+		}
+
+		// TCP DNS requests allcate a two byte length field prefixed to the message.
```
loshz, line 250:
```suggestion
		// TCP DNS requests allocate a two byte length field prefixed to the message.
```

pkg/dns/dns.go
```diff
@@ -0,0 +1,270 @@
...
+			// to close the connection
+			continue
+		}
+
+		// Now that we have the request we can forward the dnsrequest to consul
+		req := &pbdns.QueryRequest{
+			Msg:      data,
+			Protocol: pbdns.Protocol_PROTOCOL_TCP,
+		}
+		resp, err := client.Query(context.Background(), req)
```
loshz, line 235: Should the context have a timeout?

pkg/dns/dns.go
chapmanc: That's a good question. To be honest I think we should be fine since the worst that can happen is that we block on a go routine and the upper bound is just the memory of the system. I am inclined to leave it without one and see how we do. A load test would be a good call here. I'll look into doing that and see how we perform.

pkg/dns/dns_test.go
```diff
@@ -0,0 +1,282 @@
...
+		s.T().FailNow()
+	}
+	err = server.Run()
+	if err != nil {
+		s.T().FailNow()
+	}
+	server.Stop()
+
+	s.Require().Eventually(func() bool {
+		port := server.TcpPort()
```
mkeeler, line 83: I would probably grab the port before the call to `server.Stop()` As is this seems to be relying on the fact that the state of the listener will not change when the connection is closed which seems like an implementation detail we may not want to rely on.

pkg/dns/dns_test.go
```diff
@@ -0,0 +1,282 @@
...
+		addr := fmt.Sprintf("127.0.0.1:%v", port)
+		_, err := net.Dial("tcp", addr)
+		s.T().Logf("dial error: %v", err)
+		return err != nil
+	}, time.Second*20, time.Second, "Failure to shut down tcp")
+
+	s.Require().Eventually(func() bool {
+		port := server.TcpPort()
```
mkeeler, line 91:
```suggestion
		port := server.UdpPort()
```

Again, I would get the UDP port prior to the Stop call.

pkg/dns/dns_test.go
```diff
@@ -0,0 +1,282 @@
...
+		stopCh:  stopChan,
+	}
+
+	go server.proxyUDP()
+
+	testCases := map[string]struct {
+		dnsRequest   []byte
+		dnsResp      []byte
+		expected     error
+		largeResp    error
```
mkeeler, line 123:
```suggestion
```

This appears to be unused.

pkg/dns/dns_test.go
```diff
@@ -0,0 +1,282 @@
...
+		stopCh:      stopChan,
+	}
+
+	go server.proxyTCP()
+
+	testCases := map[string]struct {
+		dnsRequest   []byte
+		dnsResp      []byte
+		expected     error
+		largeResp    error
```
mkeeler, line 213: nit:

```suggestion
		largeResp    bool
```

The error value appears to never be used but instead we only ever care whether its nil or not. In that case you could have this be a bool instead. You would also need to change the two places its referenced.

pkg/dns/dns.go
mkeeler: The intended use is to proxy a single workloads DNS requests and for consul-dataplane to run alongside that workload. The request throughput should be pretty minimal.

mkeeler (commented):
The code generally looks great. Most of my comments are rather nit picky or based on paranoia of future modifications surfacing bad behavior even when that behavior definitely doesn't exist today.

pkg/consuldp/consul_dataplane.go
```diff
@@ -195,6 +186,26 @@ func (cdp *ConsulDataplane) Run(ctx context.Context) error {
...
+	dnsClientInterface := pbdns.NewDNSServiceClient(cdp.serverConn)
+
+	dnsServer, err := dns.NewDNSServer(dns.DNSServerParams{
+		BindAddr: cdp.cfg.DNSServer.BindAddr,
+		Port:     cdp.cfg.DNSServer.Port,
+		Client:   dnsClientInterface,
+		Logger:   cdp.logger,
+	})
+	if err == dns.ErrServerDisabled {
+		return nil, err
```
mkeeler, line 199:
```suggestion
		return nil, nil
```

If you do this then you wont have to check for ErrServerDisabled further up.

mkeeler (approved):
LGTM

I had a few more comments but they are extremely minor and you can take or leave them.

pkg/consuldp/consul_dataplane.go
```diff
@@ -152,13 +154,25 @@ func (cdp *ConsulDataplane) Run(ctx context.Context) error {
 		return fmt.Errorf("failed to run proxy: %w", err)
 	}
 
+	dnsServer, err := cdp.startDNSProxy(ctx)
+	if err != nil {
+		if err != dns.ErrServerDisabled {
```
riddhi89, line 159: If there is an error on start up, the envoy proxy process will linger around (as it runs as a sub-process) although the parent consul-dataplane process has died. 
It might be a good idea to start the DNS proxy before the envoy proxy so that if there is an issue on start-up, we hard fail right away before trying to start the envoy proxy as a sub-process.

riddhi89 (changes requested):

pkg/consuldp/consul_dataplane.go
chapmanc: Hmm this seems to be more of a bug in main where we don't call cancel on the context if run returns a failure. I think it would make more sense to call cancel in the error handling in main. Thoughts?

riddhi89: So far run returning a failure has been more immediate as a result of synchronous calls erroring out. 
If there is a failure in async go routines, [appropriate cleanup](https://github.com/hashicorp/consul-dataplane/blob/main/pkg/consuldp/consul_dataplane.go#L156-L172) is done before exiting. So if you are referring to calling `cancel` here - https://github.com/hashicorp/consul-dataplane/blob/main/cmd/consul-dataplane/main.go#L210, it might not work out since we would need to wait for everything to be cancelled before exiting?

pkg/consuldp/consul_dataplane.go
```diff
@@ -152,6 +156,11 @@ func (cdp *ConsulDataplane) Run(ctx context.Context) error {
 		return fmt.Errorf("failed to run proxy: %w", err)
 	}
 
+	if err = cdp.startDNSProxy(ctx); err != nil {
+		cdp.logger.Error("failed to start the dns proxy", "error", err)
+		return err
+	}
+
```
riddhi89, line 163: would be a good idea to call a `defer dnsServer.Stop()` so if Run returns ( due to proxy exiting or the xds server exiting), dns server is gracefully stopped.

pkg/consuldp/consul_dataplane.go
chapmanc: We should chat as that's not how I think we handle exiting. If there is an error the context should close and all dependent go routines should close. I did the cancel in that error handling and the behaviour you saw was fixed. 

I have tested it and it does work. The context handler stops everything and since it's a channel the communication should be async. It's the same behaviour as if you called SIGTERM.

pkg/dns/dns.go
```diff
@@ -0,0 +1,314 @@
...
+	d.connUDP = connUDP
+
+	// 2. Setup tcp listener
+	tcpAddr := &net.TCPAddr{
+		Port: d.port,
+		IP:   d.bindAddr,
+	}
+	listenerTCP, err := net.ListenTCP("tcp", tcpAddr)
+	if err != nil {
+		return fmt.Errorf("error listening to udp: %w", err)
```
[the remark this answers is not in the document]
chapmanc: Good catch :)

pkg/consuldp/consul_dataplane.go
riddhi89: Sure, lets chat. Basically, there is nothing sending a stop to the envoy proxy explicitly since it does not share the parent process context.

riddhi89: >It's the same behaviour as if you called SIGTERM.

It is slightly different because there is an async go routine listening for the cancel to handle appropriate cleanup: https://github.com/hashicorp/consul-dataplane/blob/main/pkg/consuldp/consul_dataplane.go#L156

pkg/consuldp/consul_dataplane.go
chapmanc: I've moved this ahead of the envoy start.

cmd/consul-dataplane/main.go
```diff
@@ -207,6 +218,7 @@ func main() {
 
 	err = consuldpInstance.Run(ctx)
 	if err != nil {
+		cancel()
```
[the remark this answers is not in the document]
chapmanc: We discussed and I will address this in a separate PR as a context aware in each of the go routines to signal closing.

PR #24, opened 2022-10-06
Make xdsserver and envoy context aware

chapmanc:
Making the envoy command context aware and the xdsServer so when the context is done the envoy process stops and the xdsServer stops

### DIFF
--- a/pkg/envoy/proxy.go
+++ b/pkg/envoy/proxy.go
@@ -1,6 +1,7 @@
 package envoy
 
 import (
+	"context"
 	"errors"
 	"fmt"
 	"io"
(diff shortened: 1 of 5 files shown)


pkg/envoy/proxy.go
```diff
@@ -257,7 +258,7 @@ func (p *Proxy) buildCommand(cfgPath string) *exec.Cmd {
 		p.cfg.ExtraArgs...,
 	)
 
-	cmd := exec.Command(p.cfg.ExecutablePath, args...)
+	cmd := exec.CommandContext(ctx, p.cfg.ExecutablePath, args...)
```
riddhi89, line 261: Does this cancel the context after the command completes , i.e, the envoy proxy has started?
The description here makes it seem otherwise: https://pkg.go.dev/os/exec#CommandContext

riddhi89: Nvm, based on your test below, looks like it does.

riddhi89 (approved):
LGTM

pkg/envoy/proxy.go
boxofrad: Yep I think you're right @chapmanc, the goroutine should unblock and do the cleanup. Maybe worth sticking a `fmt.Println` in there and testing it manually, though?

boxofrad (approved):
Looks good to me! Could we remove `Proxy.Stop` though, so there's just one way to shut the proxy down? 🙏🏻

pkg/envoy/proxy.go
chapmanc: Confirmed after passing a logger into the test.

PR #25, opened 2022-10-07
Make metrics server context aware and centralize into metrics config

chapmanc:
Trying to make sure everything uses the ctx that is passed down and that we exit based off if the context is done.

Also centralizing the logic for metrics inside the `metricsConfig` struct so that the `ConsuldDataplaneInstance`  doesn't get too large over time. 

This also mimics consul's metric config so we keep the code bases semi similar.


pglass (approved):

riddhi89 (approved):
LGTM

PR #26, opened 2022-10-10
ci: Fix changelog-checker GHA workflow

eculver:
This is a follow-on to https://github.com/hashicorp/consul/pull/14842. I noticed that this repo borrowed its changelog checker workflow from Consul and so I thought I'd port the fix from the aforementioned Consul PR so that this too can be fixed.

See the description in the other PR for more details.


riddhi89 (approved):
Thanks for fixing it here too!

PR #27, opened 2022-10-10
Build docker redhat UBI for dockerhub

riddhi89:
To release consul-dataplane redhat/UBI image to dockerhub via CRT workflows.


marianoasselborn:
I think this looks good @riddhi89 , I can see the new image tarball in the artifacts list and the list of tags 👍

marianoasselborn (approved):

PR #34, opened 2022-10-19
Adding statsD and dogstatsD support

chapmanc:
This PR is best viewed commit by commit. 

1. e7347d3c9ff829180a6b72c2b3984e9652075c0a adds support for statsD and dogstatsD if the envoy proxy is configured with dogstatsD or statsD and configure's the consul dataplane metrics to sink to statsD or dogtstatsD. This approach means we leave it up envoy  to push it's own metrics statsD/dogstatsD and we push our own metrics as well. service metrics would have to also push to statsD itself (not sure of this so have to sync with @pglass)
2. 7d712cdbd36fdc4ec126195c3c4da41fe48b4e7d cleaning up the logger in the stream interceptor that was used only for testing 
3. b1104c1498fd46c9a2383ce5cb8561dc6864372b was to rename functions to hopefully be a bit more clear on the purpose.

### DIFF
--- a/pkg/consuldp/metrics_test.go
+++ b/pkg/consuldp/metrics_test.go
@@ -13,16 +13,42 @@ import (
 	"testing"
 	"time"
 
+	"github.com/armon/go-metrics"
 	"github.com/hashicorp/consul-dataplane/internal/bootstrap"
 	metricscache "github.com/hashicorp/consul-dataplane/pkg/metrics-cache"
 	"github.com/stretchr/testify/require"
 )
 
 var (
+	dogStatsdAddr    = "127.0.0.1"
 	envoyMetricsPort = 19000
 	envoyMetricsAddr = "127.0.0.1"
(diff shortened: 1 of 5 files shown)


pkg/consuldp/metrics_test.go
```diff
@@ -13,17 +13,43 @@ import (
...
+}
+
+func assertServerMatchesExpected(t *testing.T, server *net.UDPConn, buf []byte, expected string) {
+	t.Helper()
+	n, _ := server.Read(buf)
+	msg := string(buf[:n])
+
+	if msg != expected {
+		t.Fatalf("Line %s does not match expected: %s", msg, expected)
+	}
```
pglass, line 50: nit: could this use `require.Equal`?

```suggestion
	require.Equal(t, expected, msg, "Line %s does not match expected: %s", msg, expected)
```

pglass (approved):

PR #39, opened 2022-10-25
Integration tests 🔬

boxofrad:
This PR adds a small set of integration tests to our build pipeline, covering the end-to-end process of creating a service mesh using Consul Dataplane, including:

* ACL Login
* gRPC TLS Configuration
* Envoy Bootstrap Configuration
* xDS Proxying (and Server xDS)
* Central Service Configuration
* L7 Intentions
* DNS Proxying (both UDP and TCP)
* Metrics Merging

It does so by running Consul Servers and Dataplanes in Docker containers (using the awesome [testcontainers](https://github.com/testcontainers/testcontainers-go) library ⭐).

The eventual aim is to run the [existing suite of integration tests](https://github.com/hashicorp/consul/tree/main/test/integration/connect/envoy) in the core Consul repository against Consul Dataplane, which has much more comprehensive coverage of our service mesh features, but would require some significant structural changes to support "agentless" deployments.


integration-tests/main_test.go
```diff
@@ -0,0 +1,211 @@
...
+
+	// disableReaper controls whether the container reaper is enabled. It can be
+	// configured using the -disable-reaper flag.
+	disableReaper bool
+
+	// upstreamLocalBindPort is the port the frontend sidecar will bind the local
+	// listener for the backend upstream to.
+	upstreamLocalBindPort = tcpPort(10000)
+
+	// proxyInboundListenerPort is the port the backend sidecar will bind its
```
riddhi89, line 40: looks like both the frontend and backend sidecars have the same public listener port (not just backend per comment).

riddhi89 (commented):
Excellent job! Appreciate how the test files and helpers have been split out and organized 👏

boxofrad:
Thanks @riddhi89 🙇🏻  I've managed to get it running in GitHub actions now, which took some reworking as Docker bind-mounts are a little tricky under SELinux.

Were there any parts you found unintuitive, that could use some explanation/doc comments?

integration-tests/main_test.go
boxofrad: Yep, you're totally right, will update! It's just that we don't ever actually use the frontend proxy's public listener port.

boxofrad:
Heads up @eculver 🙇🏻‍♂️

.github/workflows/build.yml
```diff
@@ -256,3 +256,48 @@ jobs:
           dev_tags: |
             docker.io/hashicorppreview/${{ env.repo }}:${{ env.dev_tag }}-ubi
             docker.io/hashicorppreview/${{ env.repo }}:${{ env.dev_tag }}-ubi-${{ github.sha }}
+
```
boxofrad, line 259: Hey @claire-labry and @marianoasselborn 👋🏻 

I think this part needs your review because RelEng are the codeowners of our build configuration.

I did originally have these tests in a separate workflow, but I wanted to run them against the real release Docker image because we also bundle Envoy and go-discover.

.github/workflows/build.yml
```diff
@@ -256,3 +256,48 @@ jobs:
...
+  integration-tests:
+    name: Integration Tests (Consul ${{ matrix.server.version }})
```
NiniOak, line 261: Why's the server version required in the test name? Shouldn't this be a constant?

Makefile
```diff
@@ -109,3 +109,18 @@ ifdef LAST_RELEASE_GIT_TAG
...
+integration-tests: docker/release-default
+ifdef INTEGRATION_TESTS_OUTPUT_DIR
+# make's built-in realpath function doesn't support non-existent directories
+# and intermittently has issues finding newly created ones (so preemptively
+# creating it with mkdir isn't an option) so we'll rely on the realpath bin.
+ifeq (, $(shell which realpath))
+ $(error "GNU Coreutils are required to run the integration-tests target with INTEGRATION_TESTS_OUTPUT_DIR.")
+endif
+endif
+	cd integration-tests && go test -v ./ -output-dir="$(shell realpath $(INTEGRATION_TESTS_OUTPUT_DIR))" -dataplane-image="$(INTEGRATION_TESTS_DATAPLANE_IMAGE)" -server-image="$(INTEGRATION_TESTS_SERVER_IMAGE)"
```
NiniOak, line 126: Pardon my questions, still new at the company. 

Why does the server-image need to be defined here? Can the image already be defined in the Github config.yaml be used here as well? So there's no need to change the value in more than one place for future releases.

.github/workflows/build.yml
boxofrad: Good question! Consul Dataplane is compatible with multiple versions of the Consul server, starting with 1.14, to make upgrades easy — so we'll want to run the integration test suite against each of them to check we haven't broken anything on either side.

Makefile
NiniOak: Makes sense. I wonder if there's a way to inform anyone running this locally to change server image value to whatever the latest version is/find a way to keep this value up to date. I foresee this value becoming stale quickly.

boxofrad: Definitely! I think once 1.14 ships we can just switch to the `latest` tag. Your local cache could still become stale, but it's a more sensible default, at least.

boxofrad:
@NiniOak The `_test.go` suffix ensures these helpers only get built (by the Go tooling) as part of a test binary. That said, this entire package only contains tests so it doesn't really make much difference.

Maybe it'd be clearer to put everything in a `helpers` subpackage, and import it [with `.`](https://stackoverflow.com/a/6478990) for convenience?

integration-tests/suite_test.go
```diff
@@ -0,0 +1,250 @@
...
+		if err := os.WriteFile(filepath.Join(outputDir, name), data, 0660); err != nil {
+			t.Logf("failed to write artifact %s: %v", name, err)
+		}
+	}
+}
+
+// Volume returns a Docker volume that can be used to share files between
+// containers. You can also add files from the host using WriteFile. The
+// volume will be deleted when the test finishes.
+func (s *Suite) Volume(t *testing.T) *Volume {
```
pglass, line 156: Totally non-blocking / optional: Did you consider docker compose for setting up some of this? 

I think a compose file could get us the docker network, the volume, and the pause container pretty easily, and we could launch additional containers from golang at that point. The docker compose config could also be customizable with something like [env variable substitution](https://docs.docker.com/compose/environment-variables/), an [override file](https://docs.docker.com/compose/extends/#multiple-compose-files), or maybe by templating the compose file.

Feel free to ignore this, since it would basically be a significant rewrite (and I'm probably missing context on broader integration test concerns/plans). Also, compose may not ultimately be better. I do think maintaining compose yaml for setup/teardown would be easier to grok. Although, we still need go code to inspect container logs to and wait for startup and such.

pglass (approved):
This is great! A few comments/thoughts that are not blocking.

integration-tests/helpers_test.go
```diff
@@ -0,0 +1,140 @@
...
+
+func udpPort(n int) nat.Port {
+	port, err := nat.NewPort("udp", strconv.Itoa(n))
+	if err != nil {
+		panic(err)
+	}
+	return port
+}
+
+func RegisterSytheticNode(t *testing.T, client *api.Client) {
```
nfi-hashicorp, line 35: :D
```suggestion
func RegisterSyntheticNode(t *testing.T, client *api.Client) {
```

boxofrad: Whoops! Good catch 😄

integration-tests/helpers_test.go
```diff
@@ -0,0 +1,140 @@
...
+
+	_, err := client.Catalog().Register(&api.CatalogRegistration{
+		Node:           syntheticNodeName,
+		SkipNodeUpdate: true,
+		Service:        service,
+	}, nil)
+	require.NoError(t, err)
+}
+
+func SetConfigEntry(t *testing.T, client *api.Client, entry api.ConfigEntry) {
```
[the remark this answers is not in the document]
boxofrad: Good point! I think that would look a lot cleaner 👍🏻

integration-tests/helpers_test.go
```diff
@@ -0,0 +1,140 @@
...
+		if err != nil {
+			t.Logf("HTTP access check failed: %v\n", err)
+		}
+		return ok
+	}, time.Minute, 1*time.Second)
+}
+
+func canAccess(ip string, port int) (bool, error) {
+	url := fmt.Sprintf("http://%s:%d/", ip, port)
+	rsp, err := http.Get(url)
```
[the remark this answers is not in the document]
boxofrad: Good catch, will add one in!

integration-tests/helpers_test.go
```diff
@@ -0,0 +1,140 @@
...
+
+	bytes, err := io.ReadAll(rsp.Body)
+	return false, fmt.Errorf("unexpected response status: %d - body: %s", rsp.StatusCode, bytes)
+}
+
+func DNSLookup(t *testing.T, suite *Suite, protocol string, serverIP string, serverPort int, host string) []string {
+	t.Helper()
+
+	req := new(dns.Msg)
+	req.SetQuestion(host, dns.TypeA)
```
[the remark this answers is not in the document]
boxofrad: Yeah, I went with A because it's the easiest and I don't think we need to test the DNS server comprehensively here (just check that we're properly proxying it).

integration-tests/main_test.go
```diff
@@ -0,0 +1,258 @@
...
+	serverHTTPPort = tcpPort(8500)
+
+	// dnsUDPPort is UDP the port Consul Dataplane's DNS proxy wil be bound to.
+	dnsUDPPort = udpPort(40000)
+
+	// dnsTCPPort is TCP the port Consul Dataplane's DNS proxy wil be bound to.
+	dnsTCPPort = tcpPort(40000)
+
+	// metricsPort is the port Consul Dataplane will serve merged prometheus
+	// metrics on.
```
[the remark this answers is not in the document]
boxofrad: Ah yeah, it took me a little while to grok this enough to write the test, and it looks like I never updated the comment!

integration-tests/suite_test.go
boxofrad: Great point! To be totally honest, it's been several years since I've used compose - so it isn't in my frequent grab-bag of tools 😄 

I have a vague preference for keeping as much as possible in Go, because I find that this kind of test suite can become quite hard to reason about once you're shelling-out to different tools and templating stuff. That said, I appreciate that some of the helper functions here aren't the easiest to follow!

NiniOak:
> Maybe it'd be clearer to put everything in a helpers subpackage, and import it [with .](https://stackoverflow.com/a/6478990) for convenience?

Yes, I believe using a subpackage and importing  it will be the better pattern.

PR #40, opened 2022-10-26
Set prometheus retention time

pglass:
This is a fix to pass the `-telemetry-prom-retention-time` value through to the Prometheus sink.

### DIFF
--- a/cmd/consul-dataplane/main.go
+++ b/cmd/consul-dataplane/main.go
@@ -144,6 +144,10 @@ func validateFlags() {
 	default:
 		log.Fatal("invalid log level. valid values - TRACE, DEBUG, INFO, WARN, ERROR")
 	}
+
+	if promRetentionTime <= 0 {
+		log.Fatal("Prometheus retention time must be greater than 0")
+	}
 }
 
 func main() {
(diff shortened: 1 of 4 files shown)


pkg/consuldp/metrics.go
```diff
@@ -266,6 +266,7 @@ func (m *metricsConfig) getPromDefaults() (*prom.Registry, *prometheus.Prometheu
 		return nil, nil, err
 	}
 	opts := &prometheus.PrometheusOpts{
+		Expiration:       m.cfg.Prometheus.RetentionTime,
```
pglass, line 269: This looks like the right thing based on what Consul does: https://github.com/hashicorp/consul/blob/b9f70c0bf1a6db535663158ac943542900c3c757/agent/config/builder.go#L929

cmd/consul-dataplane/main.go
```diff
@@ -107,7 +108,7 @@ func init() {
 
 	flag.BoolVar(&useCentralTelemetryConfig, "telemetry-use-central-config", true, "Controls whether the proxy applies the central telemetry configuration.")
 
-	flag.StringVar(&promRetentionTime, "telemetry-prom-retention-time", "", "The duration for Prometheus metrics aggregation.")
```
pglass, line 110: I chose a default of 60s because that is the default go-metrics uses: https://github.com/hashicorp/go-metrics/blob/master/prometheus/prometheus.go#L22 🤷

chapmanc (approved):
LGTM

PR #42, opened 2022-10-27
Default value for the prometheus scrape path

pglass:
This sets a default value for the `-telemetry-prom-scrape-path` and adds validation to ensure that it is non-empty. When the scrape path was empty by default, Envoy would be configured in way that made it impossible to retrieve metrics.


cmd/consul-dataplane/main.go
```diff
@@ -113,7 +113,7 @@ func init() {
 	flag.StringVar(&promKeyFile, "telemetry-prom-key-file", "", "The path to the client private key used to serve Prometheus metrics.")
 	flag.StringVar(&promCertFile, "telemetry-prom-cert-file", "", "The path to the client certificate used to serve Prometheus metrics.")
 	flag.StringVar(&promServiceMetricsURL, "telemetry-prom-service-metrics-url", "", "Prometheus metrics at this URL are scraped and included in Consul Dataplane's main Prometheus metrics.")
-	flag.StringVar(&promScrapePath, "telemetry-prom-scrape-path", "", "The URL path where Envoy serves Prometheus metrics.")
+	flag.StringVar(&promScrapePath, "telemetry-prom-scrape-path", "/metrics", "The URL path where Envoy serves Prometheus metrics.")
```
pglass, line 116: I chose /metrics to match the [k8s docs](https://developer.hashicorp.com/consul/docs/k8s/connect/observability/metrics)?

ishustava (approved):

PR #46, opened 2022-11-04
Binary build fixes for Windows

pglass:
* Switch bootstrap config file from a named Unix pipe to a tempfile, since there shouldn't be anything sensitive in the bootstrap config file
* Remove `x-consul-token` from the bootstrap config template since the token is injected into the gRPC stream by an interceptor

### DIFF
--- a/internal/bootstrap/bootstrap_tpl.go
+++ b/internal/bootstrap/bootstrap_tpl.go
@@ -43,12 +43,6 @@ type BootstrapTplArgs struct {
 	// service and is expected to be used for that purpose.
 	LocalAgentClusterName string
 
-	// Token is the Consul ACL token provided which is required to make gRPC
-	// discovery requests. If non-empty, this must be configured as the gRPC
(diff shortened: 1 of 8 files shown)


mkeeler (approved):

internal/bootstrap/bootstrap_tpl.go
```diff
@@ -272,12 +266,7 @@ const bootstrapTemplate = `{
       "api_type": "DELTA_GRPC",
       "transport_api_version": "V3",
       "grpc_services": {
-        "initial_metadata": [
-          {
-            "key": "x-consul-token",
-            "value": "{{ .Token }}"
-          }
-        ],
```
riddhi89, line 280: since this is generated code, worth implementing the hackery in the makefile helper : https://github.com/hashicorp/consul-dataplane/blob/main/Makefile#L80

joselo85:
Just to let you know, I am using this code (updated with main branch of course) and using it on Windows nodes in AWS, and for now it seems to work fine!

PR #47, opened 2022-11-09
Support environment variables

pglass:


boxofrad (approved):
This is awesome, nice job! 🙌🏻 

One thought for future consideration: when we add support for an HCL config file, we may want to adopt a pattern similar to Consul core where we read the different input sources (file, env, flags) independently and merge them together. It may also help with testability. Not a blocker for now, though!

curtbushko (approved):

pglass:
> One thought for future consideration: when we add support for an HCL config file, we may want to adopt a pattern similar to Consul core where we read the different input sources (file, env, flags) independently and merge them together. It may also help with testability. Not a blocker for now, though!

💯 I originally was going to add just a few one-off variables, but then it was easy to update every flag with an environment variable.

I'm going to merge this after testing on k8s: https://github.com/hashicorp/consul-k8s/pull/1676, and I'll follow up in another PR with some unit testing for this.

PR #53, opened 2022-11-16
Run integrations tests with UBI image

pglass:


riddhi89 (approved):

.github/workflows/build.yml
```diff
@@ -258,29 +258,32 @@ jobs:
             docker.io/hashicorppreview/${{ env.repo }}:${{ env.dev_tag }}-ubi-${{ github.sha }}
 
   integration-tests:
-    name: Integration Tests (Consul ${{ matrix.server.version }})
+    name: Integration Tests (Consul ${{ matrix.server.version }} ${{ matrix.dataplane.docker_target }})
     needs:
       - build-docker-default
```
boxofrad, line 263: Is it possible to parameterize the `needs` list using the matrix? For UBI I think we need to wait on the `build-docker-ubi-dockerhub` step.

pglass: Ah, good point!

It looks like matrix items can't be used in the needs list, though: https://github.com/orgs/community/discussions/25364

pglass: > It looks like matrix items can't be used in the needs list, though:

Double checked this here: https://github.com/hashicorp/consul-dataplane/commit/f73d449bb6c85b8fd639dd112af9c8fa1188e7ea 

https://github.com/hashicorp/consul-dataplane/actions/runs/3482299777

```
[Invalid workflow file: .github/workflows/build.yml#L263](https://github.com/hashicorp/consul-dataplane/actions/runs/3482299777/workflow)
The workflow is not valid. .github/workflows/build.yml (Line: 263, Col: 9): Unrecognized named-value: 'matrix'. Located at position 1 within expression: matrix.dataplane.docker_target
```

😞

PR #74, opened 2023-02-16
Update consul-server-connection-manager to 0.1.2

curtbushko:
- Consul-server-connection-manager was updated to Go 1.20.1 and there was a bug fix so pick that up.


DanStough (approved):
LGTM!

PR #90, opened 2023-03-07
[CC-4361] Add HCPMetricsBindSocketDir to Envoy bootstrap config

Achooo:
This PR is very similar to https://github.com/hashicorp/consul/pull/16511 in `consul`

## Background
The HCP Cloud team is working on enabling observability metrics in HCP. 
In particular, we want to forward Envoy metrics.

This PR aims to allow Envoy bootstrap configuration of `envoy_hcp_metrics_bind_socket_dir`, a directory where a unix socket will be created with the name `<namespace>_<proxy_id>.sock` to forward Envoy metrics.

## Code changes 
- `internal/bootstrap/bootstrap_config.go` : Add a `HCPMetricsBindSocketDir` string type in the `BootstrapConfig` struct and a function `appendHCPMetricsConfig` to generate the socket name, configure the stats sink and cluster for metrics forwarding.
- `internal/bootstrap/bootstrap_config_test.go`: add relevant tests to the above
- `pkg/consuldp/bootstrap_test.go`: Add a test to generate a golden test file
- `internal/bootstrap/bootstrap_tpl.go` : small change to move brackets `[]` into the template instead, which facilitates config generation

## Next Steps

There will be a follow-up PR to generate a listener and cluster dynamically that will be listening on this socket 
That listener and cluster will be configured to route to collector instances through the service mesh.

Caveat: we needed to add this local listener indirection as a workaround due to this [issue](https://github.com/envoyproxy/envoy/issues/25700)

### DIFF
--- a/internal/bootstrap/bootstrap_config.go
+++ b/internal/bootstrap/bootstrap_config.go
@@ -48,6 +48,11 @@ type BootstrapConfig struct {
 	// stats_config.stats_tags can be made by overriding envoy_stats_config_json.
 	StatsTags []string `mapstructure:"envoy_stats_tags"`
 
+	// HCPMetricsBindSocketDir is a string that configures the directory for a
+ 	// unix socket where Envoy will forward metrics. These metrics get pushed to
+ 	// the HCP Metrics collector to show service mesh metrics on HCP.
+ 	HCPMetricsBindSocketDir string `mapstructure:"envoy_hcp_metrics_bind_socket_dir"`
+
 	// PrometheusBindAddr configures an <ip>:<port> on which the Envoy will listen
 	// and expose a single /metrics HTTP endpoint for Prometheus to scrape. It
 	// does this by proxying that URL to the internal admin server's prometheus
@@ -237,6 +242,11 @@ func (c *BootstrapConfig) ConfigureArgs(args *BootstrapTplArgs, omitDeprecatedTa
 		args.StatsFlushInterval = c.StatsFlushInterval
 	}
 
+	// Setup HCP Metrics if needed. This MUST happen after the Static*JSON is set above
+	if c.HCPMetricsBindSocketDir != "" {
+		appendHCPMetricsConfig(args, c.HCPMetricsBindSocketDir)
+	}
+
 	return nil
 }
 
@@ -270,7 +280,7 @@ func (c *BootstrapConfig) generateStatsSinks(args *BootstrapTplArgs) error {
 	}
 
 	if len(stats_sinks) > 0 {
-		args.StatsSinksJSON = "[\n" + strings.Join(stats_sinks, ",\n") + "\n]"
+		args.StatsSinksJSON = strings.Join(stats_sinks, ",\n")
 	}
 	return nil
 }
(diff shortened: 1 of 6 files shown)


DanStough:
As we try to add support for VMs to consul-dataplane in the future, I believe using a Unix socket going to be problematic for Windows. Is this a cause for concern?

internal/bootstrap/bootstrap_config_test.go
```diff
@@ -1540,3 +1591,238 @@ func TestConsulTagSpecifiers(t *testing.T) {
...
+			inputArgs: &BootstrapTplArgs{
+				Namespace: "default",
+				ProxyID: "web-sidecar-proxy",
+			},
+			bindSocketDir: "/tmp/consul/hcp-metrics",
+			wantArgs: &BootstrapTplArgs{
+				Namespace: "default",
+				ProxyID: "web-sidecar-proxy",
+				StatsSinksJSON: `{
+					"name": "envoy.stat_sinks.metrics_service",
```
hashi-derek, line 1611: If this is repeated in a bunch of locations, maybe it's best to define it as a constant in the test somewhere.

internal/bootstrap/bootstrap_tpl.go
```diff
@@ -263,7 +263,9 @@ const bootstrapTemplate = `{
     {{- end }}
   },
   {{- if .StatsSinksJSON }}
-  "stats_sinks": {{ .StatsSinksJSON }},
+  "stats_sinks": [
+    {{ .StatsSinksJSON }}
```
hashi-derek, line 267: Rather than defining this as a single value and worrying about concatenating commas in the go code, couldn't we make this variable a slice of strings and do the `,` joining in the template?

hashi-derek (approved):
This looks good after having a discussion with Freddy about the feature. Just putting some minor nits here that you don't necessarily have to address.

internal/bootstrap/bootstrap_config.go
```diff
@@ -795,6 +805,60 @@ func (c *BootstrapConfig) generateListenerConfig(args *BootstrapTplArgs, bindAdd
...
 
+// appendHCPMetricsConfig generates config to enable a socket at path: <hcpMetricsBindSocketDir>/<namespace>_<proxy_id>.sock
+// or <hcpMetricsBindSocketDir>/<proxy_id>.sock, if namespace is empty.
+func appendHCPMetricsConfig(args *BootstrapTplArgs, hcpMetricsBindSocketDir string) {
+	dir := hcpMetricsBindSocketDir
+	if !strings.HasSuffix(dir, "/") {
+		dir += "/"
+	}
+
+	path := fmt.Sprintf("%s%s_%s.sock",dir,args.Namespace,args.ProxyID)
```
Achooo, line 816: After testing locally with @freddygv , looks like the Namespace is always `default`, so we remove the if statement.  

See the equivalent in `consul`: https://github.com/hashicorp/consul/pull/16511/files#diff-3d65fe35c13bd0d74e9cabcee963eea26abcafb433b75454c7bfd697f1f451f4R820

internal/bootstrap/bootstrap_config_test.go
Achooo: Yeah we weren't sure if inline is easier to read (gotta jump around) but I refactored it since it's a lot ! Thanks :)

internal/bootstrap/bootstrap_tpl.go
Achooo: Yeah I agree the current state isn't ideal... I'd prefer to park this scope-wise!

I'd have to convert the actual BootstrapConfig `StatsConfigJSON` to a list due to these lines
https://github.com/hashicorp/consul-dataplane/blob/dc0da3cd4ecd2a9e15ed462f5844d7e356ab037d/internal/bootstrap/bootstrap_config.go#L194-L196, where the args have the actual config reconciled. This change would also affect most test and quite a few areas of the code.

My initial thought seeing this package was maybe there might be a better cleanup that could be done here using marshalling/unmarshalling, although this introduces with defining structs/models, it might reduce the code logic complexity and cleanup a most of these handlings in the code

PR #97, opened 2023-05-05
security: update Envoy to 1.26.1

DanStough:
Updating `main` to point to the latest release of Envoy, even though there may be newer versions before the 1.2.0 release. About to submit PRs for the existing release branches.

### DIFF
--- a/.changelog/97.txt
+++ b/.changelog/97.txt
@@ -1,0 +1,3 @@
+```release-note:security
+Update to Go 1.20.4 and Envoy 1.26.1 within the Dockerfile. 
+```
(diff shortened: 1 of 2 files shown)


DanStough:
Go was updated to 1.20.4 in CI as part of #94

hashi-derek (approved):
This looks okay, but we should probably be careful to ensure that we won't be affected by any of the breaking changes here: https://www.envoyproxy.io/docs/envoy/v1.26.1/version_history/v1.26/v1.26.0

DanStough:
Good callout. I will make sure we update this in Consul core for 1.16 and get some integration tests on it.

PR #99, opened 2023-05-05
security: update Envoy to 1.24.7

DanStough:
Updating the 1.1.x release branch with the latest patch of Envoy. The Golang version is also updated according to the existing root .go-version file.

### DIFF
--- a/.github/workflows/consul-dataplane-checks.yaml
+++ b/.github/workflows/consul-dataplane-checks.yaml
@@ -3,7 +3,7 @@ name: consul-dataplane-checks
 on:
   push:
     branches:
-      main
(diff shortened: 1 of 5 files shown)


.github/workflows/consul-dataplane-checks.yaml
```diff
@@ -3,7 +3,7 @@ name: consul-dataplane-checks
 on:
   push:
     branches:
-      main
+    - main
```
DanStough, line 6: VS code complains about the style here 🤷

loshz (approved):

hashi-derek (approved):

PR #100, opened 2023-05-05
cmd: add CLI flags for proxy shutdown lifecycle management

mikemorris:
Adds the following CLI flags for proxy shutdown lifecycle management, disabled by default:
-  `-shutdown-drain-listeners`
- `-shutdown-grace-period`

Adds the following CLI flags for managing where the proxy lifecycle HTTP APIs are exposed:
- `-graceful-shutdown-path`
- `-graceful-port`


cmd/consul-dataplane/main.go
```diff
@@ -73,6 +73,11 @@ var (
 
 	consulDNSBindAddr string
 	consulDNSPort     int
+
+	shutdownDrainListeners bool
+	shutdownGracePeriod    int
+	gracefulShutdownPath   string
+	gracefulPort           int
```
nathancoleman, line 80: Should this be `gracefulShutdownPort` for consistency?

mikemorris: Probably not because it will eventually be used for `POST /graceful_startup` from the post-start hook too, to block application container start until the Envoy proxy is ready.

huikang:
@mikemorris , thanks for making the PR. I am new to the world of consul-dataplane. Out of curiosity, do we have similar proxy shutdown feature in consul core (`consul connect envoy`)?

mikemorris:
> do we have similar proxy shutdown feature in consul core?

Nope, this has been a longstanding feature/fix request that we're looking to address now, added relevant issue references to original post for context. The eventual implementation of graceful shutdown in consul-dataplane will be triggered by a preStop hook in consul-k8s.

pglass (approved):

PR #101, opened 2023-05-09
Sync bootstrap config from consul/main

freddygv:
Updated with `make copy-bootstrap-config` to sync changes from https://github.com/hashicorp/consul/pull/17204

------------

UNIX domain socket paths are limited to 104-108 characters depending on the OS. This limit was quite easy to exceed when testing the feature on Kubernetes, due to how proxy IDs encode the Pod ID eg:
`metrics-collector-59467bcb9b-fkkzl-hcp-metrics-collector-sidecar-proxy`

To ensure we stay under the lower 104 character limit this commit makes a couple changes:
- Use a b64 encoded SHA1 hash of the namespace + proxy ID to create a short and deterministic socket file name.
- Add validation to proxy registrations and proxy-defaults to enforce a limit on the socket directory length.


internal/bootstrap/bootstrap_config_test.go
```diff
@@ -91,19 +94,6 @@ const (
     ]
   }
```
ndhanushkodi, line 95: just checking, is this file diff bigger because the changes weren't synced recently?

ndhanushkodi (approved):
nice!

internal/bootstrap/bootstrap_config_test.go
freddygv: It's because the original changes from #90 weren't synced via the Makefile target, and were a bit different from the Consul changes

Now things are the same on both sides

PR #106, opened 2023-05-12
Support HOST_IP expansion

### LINKED ISSUE #62: mesh-gateway doesn't support expansion of the environment variable HOST_IP
We are attempting to upgrade to the latest version of consul-k8s that uses the consul-dataplane image for mesh-gateways. The new mesh-gateways are unable to start because they're not able to parse our `envoy_dogstatsd_url`. 

Error message:
```
failed to parse address udp://${HOST_IP}:8125
```

Our consul proxy-defaults config entry specifies:
```
envoy_dogstatsd_url   = "udp://${HOST_IP}:8125"
```

Per the documentation, expansion of this variable should be supported for `envoy_dogstatsd_url` and `envoy_statsd_url`: https://developer.hashicorp.com/consul/docs/connect/proxies/envoy#envoy_statsd_url

If I turn on debug logging, I can see that the envoy configuration is properly expanding the `HOST_IP` to the correct ip address:

```
  \"stats_sinks\": [
    {
      \"name\": \"envoy.stat_sinks.dog_statsd\",
      \"typedConfig\": {
        \"@type\": \"type.googleapis.com/envoy.config.metrics.v3.DogStatsdSink\",
        \"address\": {
          \"socket_address\": {
            \"address\": \"10.0.123.456\",
            \"port_value\": 8125
          }
        }
      }
    }
  ],
```

Here is where the error is being thrown: https://github.com/hashicorp/consul-dataplane/blob/main/pkg/consuldp/metrics.go#L359-L363

And here we can see that the address being passed to `parseSinkAddr` 
https://github.com/hashicorp/consul-dataplane/blob/main/pkg/consuldp/metrics.go#L152 is the raw value from the proxy-config, 
without the variable expansion.

The relevant path expansion takes place here for the envoy bootstrap config: https://github.com/hashicorp/consul-dataplane/blob/main/internal/bootstrap/bootstrap_config.go#L278-L284

I'm not too familiar with the new dataplane functionality and the distinction between the stats sync in the envoy config and the sink being setup in `metrics.go`. If the solution is as easy as copying/moving the variable expansion code into metrics, I'm happy to do that. But I'd like to ensure this is the right path first.

---
Note that the mesh-gateway deployment in the consul-k8s helm chart no longer provides the `HOST_IP` environment variable. The behavior above exists even if I manually update the deployment to add the variable, and I plan to submit a separate PR to that project to add the variable back. Of course this isn't directly relevant to the consul-dataplane project, but I wanted to be clear that the problem isn't simply that `HOST_IP` is unset.

woz5999:
Closes #62 

I'm not sure about docs or changelog. Let me know how I can help there.

### DIFF
--- a/.changelog/106.txt
+++ b/.changelog/106.txt
@@ -1,0 +1,3 @@
+```release-note:bug
+Fix a bug that threw an error when trying to use `$HOST_IP` with metrics URLs.
+```
--- a/pkg/consuldp/metrics.go
+++ b/pkg/consuldp/metrics.go
@@ -164,14 +164,8 @@ func (m *metricsConfig) startMetrics(ctx context.Context, bcfg *bootstrap.Bootst
 			// 4. Start prometheus metrics sink
 			go m.startPrometheusMergedMetricsSink()
 		}
-		if bcfg.StatsdURL != "" {
-			url = bcfg.StatsdURL
-			if len(url) > 2 && url[0] == '$' {
-				url = os.Getenv(url[1:]
-			} else {
-				url = os.Expand(url, statsSinkEnvMapping)
-			}		 
-			addr, err := parseSinkAddr(url, Statsd)
+		if bcfg.StatsdURL != "" {	 
+			addr, err := parseSinkAddr(bcfg.StatsdURL, Statsd)
 			if err != nil {
 				return err
 			}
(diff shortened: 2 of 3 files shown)


david-yu:
Thank you we will take a look!

DanStough (changes requested):
Looks great 👌! Thank you very much for your contribution.

Only one minor comment. The most important thing would be to add a changelog. If you could create a file `.changelog/106.txt`, you can use this as a guideline:

````
```release-note:bug
Fix a bug that threw an error when trying to use `$HOST_IP` with metrics URLs.
```
````
We are good on doc since you pointed out in the original issue that our docs say this should be supported.

woz5999:
All changes should be incorporated now.

DanStough (approved):
🚀  I will merge after the tests are complete. Thank you again 🙇

PR #115, opened 2023-05-15
pkg/consuldp: add HTTP Server with endpoints for proxy lifecycle shutdown

mikemorris:


pkg/consuldp/consul_dataplane.go
```diff
@@ -217,12 +226,25 @@ func (cdp *ConsulDataplane) Run(ctx context.Context) error {
...
 			doneCh <- errors.New("metrics server exited unexpectedly")
+		case <-cdp.lifecycleConfig.lifecycleServerExited():
+			// Initiate graceful shutdown of Envoy, kill if error
+			if err := proxy.Quit(); err != nil {
+				cdp.logger.Error("failed to stop proxy", "error", err)
+				if err := proxy.Kill(); err != nil {
+					cdp.logger.Error("failed to kill proxy", "error", err)
+				}
+			}
+			doneCh <- errors.New("proxy lifecycle maangement server exited unexpectedly")
```
mikemorris, line 247: This should be the only change to existing logic aside from introducing the new proxy lifecycle management server exposing the `/shutdown` endpoint - attempting to [cleanly exit](https://www.envoyproxy.io/docs/envoy/latest/operations/admin#post--quitquitquit) the Envoy process first instead of the prior hard kill.

cmd/consul-dataplane/main.go
```diff
@@ -131,6 +133,8 @@ func init() {
 	StringVar(&readyBindAddr, "envoy-ready-bind-address", "", "DP_ENVOY_READY_BIND_ADDRESS", "The address on which Envoy's readiness probe is available.")
 	IntVar(&readyBindPort, "envoy-ready-bind-port", 0, "DP_ENVOY_READY_BIND_PORT", "The port on which Envoy's readiness probe is available.")
 	IntVar(&envoyConcurrency, "envoy-concurrency", 2, "DP_ENVOY_CONCURRENCY", "The number of worker threads that Envoy uses.")
+	IntVar(&envoyDrainTime, "envoy-drain-time", 30, "DP_ENVOY_DRAIN_TIME", "The time in seconds for which Envoy will drain connections.")
+	StringVar(&envoyDrainStrategy, "envoy-drain-strategy", "immediate", "DP_ENVOY_DRAIN_STRATEGY", "The behaviour of Envoy during the drain sequence. Determines whether all open connections should be encouraged to drain immediately or to increase the percentage gradually as the drain time elapses.")
```
mikemorris, line 137: Set [reasonable defaults for service-to-service deployments](https://www.envoyproxy.io/docs/envoy/latest/operations/cli#cmdoption-drain-time-s), and allow user configuration if needed. This is net new functionality as the prior proxy manager logic would just immediately kill the Envoy process, so the stock defaults (10 minute drain time, `"gradual"` drain strategy) would never have had any effect.

pkg/consuldp/lifecycle.go
```diff
@@ -0,0 +1,195 @@
...
+
+const (
+	// defaultLifecycleBindPort is the port which will serve the proxy lifecycle HTTP
+	// endpoints on the loopback interface.
+	defaultLifecycleBindPort = "20300"
+	cdpLifecycleBindAddr     = "127.0.0.1"
+	cdpLifecycleUrl          = "http://" + cdpLifecycleBindAddr
+
+	defaultLifecycleShutdownPath = "/shutdown"
+)
```
mikemorris, line 27: I _think_ these constants shouldn't actually be needed and I have a bit of a config plumbing thing that needs fixing, because these defaults _should_ be coming from the the CLI configuration defaults in `cmd/consul-dataplane/main.go`

pkg/consuldp/lifecycle_test.go
```diff
@@ -0,0 +1,195 @@
...
+			// Check lifecycle server graceful shutdown path configuration
+			if c.gracefulShutdownPath != "" {
+				require.Equal(t, m.gracefulShutdownPath, c.gracefulShutdownPath, "failed to set lifecycle server graceful shutdown HTTP endpoint path")
+			}
+
+			// Check lifecycle server graceful shutdown path configuration
+			url := fmt.Sprintf("http://127.0.0.1:%d%s", port, m.gracefulShutdownPath)
+			log.Printf("sending request to %s\n", url)
+
+			resp, err := http.Get(url)
```
mikemorris, line 149: Could possibly add some way to check that the Envoy process stays alive for `c.shutdownGracePeriod` seconds here.

nathancoleman (changes requested):

pkg/consuldp/consul_dataplane.go
```diff
@@ -250,20 +272,37 @@ func (cdp *ConsulDataplane) startDNSProxy(ctx context.Context) error {
...
+
+		// If not found, append value from consul-dataplane Envoy config to extraArgs
+		if setEnvoyArg {
+			extraArgs = append(extraArgs, fmt.Sprintf("%s %v", envoyArg, cdpEnvoyValue))
+		}
 	}
 
 	return envoy.ProxyConfig{
+		AdminAddr:       cdp.cfg.Envoy.AdminBindAddress,
+		AdminBindPort:   cdp.cfg.Envoy.AdminBindPort,
```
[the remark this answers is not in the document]
mikemorris: It felt reasonable to pull these out of the Envoy JSON config at this point (happy to change if they're already somewhere else I didn't notice) to make them accessible within `pkg/envoy/proxy.go` where they're needed for the HTTP calls to Envoy's admin API for the `Drain()` and `Quit()` methods.

This was not needed previously, as the Envoy process was just terminated with a process kill signal.

pkg/consuldp/config.go
```diff
@@ -274,7 +274,20 @@ type EnvoyConfig struct {
...
 	EnvoyConcurrency int
-	// ShutdownDrainListeners configures whether to wait for all proxy listeners to drain before terminating the proxy container.
+	// EnvoyDrainTime is the time in seconds for which Envoy will drain connections
+	// during a hot restart, when listeners are modified or removed via LDS, or when
+	// initiated manually via a request to the Envoy admin API.
+	// The Envoy HTTP connection manager filter will add “Connection: close” to HTTP1
+	// requests, send HTTP2 GOAWAY, and terminate connections on request completion
+	// (after the delayed close period).
+	// https://www.envoyproxy.io/docs/envoy/latest/operations/cli#cmdoption-drain-time-s
+	EnvoyDrainTime int
```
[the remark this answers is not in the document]
mikemorris: I think it would be slightly awkward as this isn't being using in Go-native functionality, it's just passed along to the Envoy binary with its' `--drain-time-s` CLI flag. Would changing this to `EnvoyDrainTimeSeconds` and the corresponding CLI arg to `-envoy-drain-time-s` be reasonable?

.github/workflows/consul-dataplane-checks.yaml
```diff
@@ -32,7 +32,7 @@ jobs:
       - uses: actions/setup-go@4d34df0c2316fe8122ab82dc22947d607c0c91f9 # v4.0.0
         with:
           go-version: ${{ needs.get-go-version.outputs.go-version }}
-      - run: go test ./...
+      - run: go test ./... -p 1 # disable parallelism to avoid port conflicts from default metrics and lifecycle server configuration
```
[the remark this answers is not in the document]
mikemorris: I'd like to clean this up, but skipped for now in the interest of expediency. It didn't feel substantial enough to warrant the effort at this time, as the full suite still completes in under a minute.

pkg/consuldp/lifecycle.go
```diff
@@ -0,0 +1,195 @@
...
+
+// lifecycleServerExited is used to signal that the lifecycle server
+// recieved a signal to initiate shutdown.
+func (m *lifecycleConfig) lifecycleServerExited() <-chan struct{} {
+	return m.errorExitCh
+}
+
+// gracefulShutdown blocks until shutdownGracePeriod seconds have elapsed, and, if
+// configured, will drain inbound connections to Envoy listeners during that time.
+func (m *lifecycleConfig) gracefulShutdown(rw http.ResponseWriter, _ *http.Request) {
```
pglass, line 149: Q: Just to think through it, what if this is called multiple times concurrently?

I think from what I can tell it's okay. And we probably aren't too concerned about concurrent local requests to this endpoint anyway.

pglass (approved):
Looks good, and this is super useful! I added some nits and non-blocking questions.

nathancoleman (approved):

pkg/consuldp/lifecycle.go
```diff
@@ -0,0 +1,195 @@
...
+)
+
+// lifecycleConfig handles all configuration related to managing the Envoy proxy
+// lifecycle, including exposing management controls via an HTTP server.
+type lifecycleConfig struct {
+	logger hclog.Logger
+
+	// consuldp proxy lifecycle management config
+	shutdownDrainListeners bool
+	shutdownGracePeriod    int
```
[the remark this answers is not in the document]
curtbushko: I agree, since it's now envoyDrainTimeSeconds.  The same change would need to happen in main.go for the flags also.

pkg/consuldp/lifecycle.go
curtbushko: Hooks in kubernetes can be called several times so we should definitely think about that scenario a little bit. The same will be true of the graceful startup side of things.

curtbushko (changes requested):
Looking pretty good! I can tell that you put a lot of thought and effort into this!

I only have a few comments, but nothing that I will stop you on.

Note: There is going to be another release of consul-dataplane tomorrow (Thursday, June 1st) so I am going to block here so this doesn't get merged in by accident.

cmd/consul-dataplane/main.go
mikemorris: The only place I see `--drain-time-s` in non-test code in Consul is as an assumption that it's used to configure hot restarts, and to disallow such configuration (it can be used for hot restarts, but is also applicable to graceful shutdown behavior as implemented here).

Similar to https://github.com/hashicorp/consul-dataplane/pull/100#issuecomment-1542330523, this behavior appears to be net-new for k8s and does not have equivalent functionality in Consul core.

pkg/consuldp/lifecycle.go
```diff
@@ -0,0 +1,195 @@
...
+		m.logger.Info("shutdown grace period timeout reached")
+		err := m.proxy.Quit()
+		if err != nil {
+			m.logger.Warn("error while shutting down Envoy", "error", err)
+			close(m.errorExitCh)
+		}
+	}()
+
+	// Wait for context timeout to elapse
+	wg.Wait()
```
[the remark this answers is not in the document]
mikemorris: This method is subsequently broken up a bit in https://github.com/hashicorp/consul-dataplane/pull/130 to allow calling the blocking function directly instead of only from the HTTP endpoint.

I don't want to change the ordering here for that reason, but would sticking the blocking call into a new goroutine make sense?
```
func (m *lifecycleConfig) gracefulShutdownHandler(rw http.ResponseWriter, _ *http.Request) {
	go m.gracefulShutdown()

	// Return HTTP 200 Success
	rw.WriteHeader(http.StatusOK)
}
```

pkg/consuldp/lifecycle.go
mikemorris: This _should_ be safe as both the `/drain_listeners` and `/quitquitquit` endpoints in Envoy are idempotent:
- https://github.com/envoyproxy/envoy/blob/3b7328f253c9cdd82dc81b736a85aeaa622262dd/source/server/admin/listeners_handler.cc#L26-L27
- https://github.com/envoyproxy/envoy/blob/3b7328f253c9cdd82dc81b736a85aeaa622262dd/source/server/admin/server_cmd_handler.cc#L24
- https://github.com/envoyproxy/envoy/blob/3b7328f253c9cdd82dc81b736a85aeaa622262dd/source/server/server.cc#L1006

Double-entry should be additionally protected by the state machine transition guards in `Drain()` and `Stop()` implemented here.

Would it be alright to plan on scoping multiple call tests to future work for job handling which will add the `preStop` hook?

mikemorris:
Addressed feedback and pushed some naming changes up into #100, which I'd to merge before this to get the chain started.

pkg/consuldp/lifecycle.go
pglass: Sounds good to me!

pglass (approved):

PR #125, opened 2023-05-30
build(deps): update UBI base to 9.2

DanStough:
Update UBI base image to 9.2 which is has and active support window.


david-yu (approved):

curtbushko (approved):
LGTM!

PR #130, opened 2023-05-30
proxy-lifecycle: catch SIGTERM and initiate graceful shutdown

mikemorris:
Refs https://github.com/hashicorp/consul-k8s/issues/536, https://github.com/hashicorp/consul-k8s/issues/650

Changes the SIGINT and SIGTERM handling logic in consul-dataplane to initiate a graceful shutdown of the managed Envoy proxy if configured. Most of the actual implementation of the graceful shutdown logic is in #115 - this PR focuses exclusively  on integrating it into the signal handling and confirming expected behavior through enhancements to the integration test.

Changes should be somewhat reasonable to review by commit if desired.

### Links
- https://golang.testcontainers.org/
- https://pkg.go.dev/github.com/testcontainers/testcontainers-go#Container
- https://pkg.go.dev/github.com/docker/docker/client#NewClientWithOpts
- https://pkg.go.dev/github.com/docker/docker/client#Client.ContainerKill
- https://pkg.go.dev/github.com/docker/docker/client#Client.ContainerStop
- https://pkg.go.dev/os/signal

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -110,7 +110,7 @@ else
 	$(error Cannot generate changelog without LAST_RELEASE_GIT_TAG)
 endif
 
-INTEGRATION_TESTS_SERVER_IMAGE    ?= hashicorppreview/consul:1.14-dev
+INTEGRATION_TESTS_SERVER_IMAGE    ?= hashicorppreview/consul:1.15-dev
 INTEGRATION_TESTS_DATAPLANE_IMAGE ?= $(PRODUCT_NAME)/release-default:$(VERSION)
 
 .PHONY: expand-integration-tests-output-dir
--- a/cmd/consul-dataplane/main.go
+++ b/cmd/consul-dataplane/main.go
@@ -80,6 +80,8 @@ var (
 	shutdownGracePeriodSeconds    int
 	gracefulShutdownPath          string
 	gracefulPort                  int
+
+	dumpEnvoyConfigOnExitEnabled bool
 )
 
 func init() {
@@ -157,6 +159,9 @@ func init() {
 	IntVar(&shutdownGracePeriodSeconds, "shutdown-grace-period-seconds", 0, "DP_SHUTDOWN_GRACE_PERIOD_SECONDS", "Amount of time to wait after receiving a SIGTERM signal before terminating the proxy.")
 	StringVar(&gracefulShutdownPath, "graceful-shutdown-path", "/graceful_shutdown", "DP_GRACEFUL_SHUTDOWN_PATH", "An HTTP path to serve the graceful shutdown endpoint.")
 	IntVar(&gracefulPort, "graceful-port", 20300, "DP_GRACEFUL_PORT", "A port to serve HTTP endpoints for graceful shutdown.")
+
+	// Default is false, may be useful for debugging unexpected termination.
+	BoolVar(&dumpEnvoyConfigOnExitEnabled, "dump-envoy-config-on-exit", false, "DP_DUMP_ENVOY_CONFIG_ON_EXIT", "Call the Envoy /config_dump endpoint during consul-dataplane controlled shutdown.")
 }
 
 // validateFlags performs semantic validation of the flag values
@@ -245,6 +250,7 @@ func main() {
 			ShutdownGracePeriodSeconds:    shutdownGracePeriodSeconds,
 			GracefulShutdownPath:          gracefulShutdownPath,
 			GracefulPort:                  gracefulPort,
+			DumpEnvoyConfigOnExitEnabled:  dumpEnvoyConfigOnExitEnabled,
 			ExtraArgs:                     flag.Args(),
 		},
 		XDSServer: &consuldp.XDSServer{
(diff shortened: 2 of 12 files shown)


pkg/consuldp/lifecycle.go
```diff
@@ -144,9 +144,18 @@ func (m *lifecycleConfig) lifecycleServerExited() <-chan struct{} {
...
 
-// gracefulShutdown blocks until shutdownGracePeriodSeconds seconds have elapsed, and, if
+func (m *lifecycleConfig) gracefulShutdownHandler(rw http.ResponseWriter, _ *http.Request) {
+	// Kick off graceful shutdown in a separate goroutine to avoid blocking
+	// sending an HTTP response
+	go m.gracefulShutdown()
+
+	// Return HTTP 200 Success
+	rw.WriteHeader(http.StatusOK)
+}
```
mikemorris, line 154: @pglass Does this feel adequate to address https://github.com/hashicorp/consul-dataplane/pull/115#discussion_r1212254465?

integration-tests/helpers/dataplane.go
```diff
@@ -70,6 +81,8 @@ func RunDataplane(t *testing.T, pod *Pod, suite *Suite, cfg DataplaneConfig) *Co
 			pod.MappedPorts[EnvoyAdminPort],
 		)
 
+		// TODO: This will fail because the integration test now performs a
+		// graceful shutdown of Envoy before reaching this point.
```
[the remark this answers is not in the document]
pglass: The purpose of this, I think, is just to dump the envoy config for debugging, particularly in the case of a test failure?

One thought is to spawn a goroutine that periodically dumps the envoy config to a file? It would do more fetches and writes than necessary, but you'd at least have the latest version of the envoy config that it was able to fetch before the Envoy admin API became unavailable.

Better would be to somehow detect that Envoy is about to be (gracefully) shutdown, and dump the Envoy config at that point. But not sure how to easily accomplish that in this situation, and it may not be worth it.

integration-tests/main_test.go
```diff
@@ -261,4 +288,85 @@ func TestIntegration(t *testing.T) {
 			return strings.Contains(output, "{\"custom_field_path\":\"/clusters\"}")
 		}, 30*time.Second, 3*time.Second, "could not find admin access logs in output")
 	}
+
+	// Overwrite deny intetion and allow two-way connections to prepare for
```
pglass, line 292:
```suggestion
	// Overwrite deny intention and allow two-way connections to prepare for
```

integration-tests/main_test.go
```diff
@@ -134,6 +137,14 @@ func TestIntegration(t *testing.T) {
 		Proxy: &api.AgentServiceConnectProxyConfig{
 			DestinationServiceName: "backend",
 			LocalServicePort:       8080,
+			Upstreams: []api.Upstream{
+				{
+					DestinationType:  api.UpstreamDestTypeService,
+					DestinationName:  "frontend",
+					LocalBindPort:    upstreamLocalBindPort.Int(),
+					LocalBindAddress: "0.0.0.0",
+				},
```
pglass, line 146: Q: Just confirming, we configure an upstream cycle (frontend and backend have each other as an upstream) in order to validate graceful shutdown when the proxy has both inbound and outbound connections?

pkg/consuldp/lifecycle.go
pglass: Yeah, I'm fine with this given your explanation there 🙂

integration-tests/main_test.go
mikemorris: Yep exactly - the basic setup was sufficient for testing that _outbound_ requests from the client to server were able to succeed during a configured shutdown grace period, but I wanted to add test coverage to ensure that _inbound_ connection (from server to client) are rejected (if `-shutdown-drain-listeners` is configured) while the client is gracefully shutting down.

integration-tests/helpers/dataplane.go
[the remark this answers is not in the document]
mikemorris: Added a proposed implementation of this in 476db3bf98497344179066b28bcc2df777b09bab

cmd/consul-dataplane/main.go
```diff
@@ -263,18 +263,20 @@ func main() {
...
 		<-sigCh
-		cancel()
+
+		consuldpInstance.GracefulShutdown(cancel)
 	}()
 
 	err = consuldpInstance.Run(ctx)
 	if err != nil {
-		cancel()
 		log.Fatal(err)
```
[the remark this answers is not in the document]
mikemorris: Implemented in c213febea2613f1ea3c2e16699991af6f32269bd

curtbushko (approved):
Looks great! 

Paul is the pro at this and I am not seeing anything other than maybe the docker image should be 1.16-dev instead of 1.15-dev.

Makefile
```diff
@@ -110,7 +110,7 @@ else
 	$(error Cannot generate changelog without LAST_RELEASE_GIT_TAG)
 endif
 
-INTEGRATION_TESTS_SERVER_IMAGE    ?= hashicorppreview/consul:1.14-dev
+INTEGRATION_TESTS_SERVER_IMAGE    ?= hashicorppreview/consul:1.15-dev
```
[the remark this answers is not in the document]
mikemorris: 1.16-dev is checked in CI, this bump was just to get past some expected CLI flags that 1.14 didn't support.

PR #133, opened 2023-06-01
fix: log level present in extra args using envoy-extra-args annotation

absolutelightning:


pkg/envoy/proxy.go
```diff
@@ -275,3 +282,14 @@ func (p *Proxy) buildCommand(ctx context.Context, cfgPath string) *exec.Cmd {
 
 	return cmd
 }
+
+// RemoveArgAndGetValue Function to get new args after removing given key
+// and also returns the value of key
+func RemoveArgAndGetValue(stringAr []string, key string) ([]string, string) {
```
Ganeshrockz, line 288: Should this be a private function instead?

pkg/envoy/proxy.go
```diff
@@ -275,3 +282,14 @@ func (p *Proxy) buildCommand(ctx context.Context, cfgPath string) *exec.Cmd {
...
+	for index, value := range stringAr {
+		if value == key {
```
Ganeshrockz, line 290: Do we need to worry about case sensitivity in this check?

pkg/envoy/proxy.go
```diff
@@ -257,6 +257,13 @@ func (p *Proxy) buildCommand(ctx context.Context, cfgPath string) *exec.Cmd {
 		logLevel = "info"
 	}
 
+	// Updating loglevel value if --log-level is present in extra args
+	newExtraArgs, valOfLoggerInExtraArgs := RemoveArgAndGetValue(p.cfg.ExtraArgs, "--log-level")
+
```
Ganeshrockz, line 262: nit: New line can be removed

pkg/envoy/proxy.go
```diff
@@ -275,3 +282,14 @@ func (p *Proxy) buildCommand(ctx context.Context, cfgPath string) *exec.Cmd {
 
 	return cmd
 }
+
+// RemoveArgAndGetValue Function to get new args after removing given key
+// and also returns the value of key
+func RemoveArgAndGetValue(stringAr []string, key string) ([]string, string) {
+	for index, value := range stringAr {
+		if value == key {
+			return append(stringAr[:index], stringAr[index+2:]...), stringAr[index+1]
```
Ganeshrockz, line 291: Wouldn't `index+2` lead to index out of range panics here?

absolutelightning: no. https://go.dev/play/p/BhUVn3iaa1K

pkg/envoy/proxy.go
absolutelightning: i guess not.

pkg/envoy/proxy.go
absolutelightning: ok yes, can be private

pkg/envoy/proxy.go
Ganeshrockz: Sorry, I also meant `index+1` in this case. Isn't this a possibility? https://go.dev/play/p/XZXT-RAHzTz

absolutelightning: log level should always have a value if key is provided.

Ganeshrockz (approved):

Ganeshrockz:
LGTM on a high level. I'd defer the approval to someone who has more context around this

mikemorris:
@absolutelightning I just merged #115 which refactored the Envoy `extraArgs` handling and I believe will conflict with this - happy to prioritize reviewing this if you can rebase on top of that!

pkg/envoy/proxy_test.go
```diff
@@ -146,3 +146,129 @@ func testOutputPath() string {
 		fmt.Sprintf("test-output-%x.json", time.Now().UnixNano()+int64(os.Getpid())),
 	)
 }
+
+func TestProxy_OverridingLoggerAndExtraArgs(t *testing.T) {
```
curtbushko, line 150: Do these tests run for you locally?  They fail for me with `failed to unmarshal output file: invalid character '\r' in string literal`

(Also, CI is not running the unit tests)

curtbushko (changes requested):
Looks pretty good.  As Mike said in another comment, you will need to rebase off of main and change several things as proxy config changed.

curtbushko:
To save you time in the future, most of use [Kind](https://kind.sigs.k8s.io/) for local testing. It runs locally and it is very nice to be able to delete and recreate your cluster very quickly. I barely ever use EKS/GKE/AKS in my testing unless I am specifically testing a feature needed on those Kubernetes versions.

Here is my demo repo and a [Makefile](https://github.com/curtbushko/cni-demo/blob/main/consul-on-kind/Makefile) that I use for quickly setting things up. Copy and replace with your directories/dockerhub. I sometimes run `make delete-cluster create-cluster` 100s of times a week. 😄

pkg/envoy/proxy_test.go
absolutelightning: @curtbushko - I tried running my test on local - test is passing but this log line is printing. 
If I run `TestProxy` I get the same logs. How should I go about fixing `TestProxy` and my tests.
Also how to make. it running in CI?

absolutelightning: Now it is running fine -

absolutelightning:
> To save you time in the future, most of use [Kind](https://kind.sigs.k8s.io/) for local testing. It runs locally and it is very nice to be able to delete and recreate your cluster very quickly. I barely ever use EKS/GKE/AKS in my testing unless I am specifically testing a feature needed on those Kubernetes versions.
> 
> Here is my demo repo and a [Makefile](https://github.com/curtbushko/cni-demo/blob/main/consul-on-kind/Makefile) that I use for quickly setting things up. Copy and replace with your directories/dockerhub. I sometimes run `make delete-cluster create-cluster` 100s of times a week. 😄

Thanks @curtbushko . Really helpful.

absolutelightning:
Also I have rebased my branch.

pkg/envoy/proxy_test.go
curtbushko: I figured out what the problem was for me. I had the `brew install base64` instead of the GNU `brew install gbase64`. This was causing `fake-envoy` to not run at all....

curtbushko:
I built my own version of the consul-dataplane image and I am running the acceptance tests against it [here](https://github.com/hashicorp/consul-k8s-workflows/actions/runs/5207404191)

curtbushko (approved):

PR #152, opened 2023-06-14
Revert Proxy Lifecycle PRs

DanStough:
Reverts:
* #130 
* #115 
* #100 
These will get backported again after some bug fixes at a later date.


curtbushko (approved):
LGTM!

PR #154, opened 2023-06-16
Add CRT updates and FIPS envoy image

skpratt:
As part of the release process for consul and consul-k8s earlier this week, CRT arrived at the conclusion that a paradigm using separate targets and container repositories for FIPS versions of Consul products was the preferred way to ensure that FIPS and non-FIPS products do not risk overwriting each other in the pipeline. This PR incorporates those changes for Dataplane, similar to what was done in the k8s PR, as well as enterprise (not linked here):
https://github.com/hashicorp/consul-k8s/pull/2333

This PR also updates the FIPS target in Dataplane to use a FIPS Envoy image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -11,6 +11,8 @@
 # prebuilt binaries in any other form.
 FROM envoyproxy/envoy-distroless:v1.26.2 as envoy-binary
 
+FROM hashicorp/envoy-fips:v1.26.2 as envoy-fips-binary
+
 # go-discover builds the discover binary (which we don't currently publish
 # either).
 FROM golang:1.20.4-alpine as go-discover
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -51,6 +53,36 @@ USER 100
 
 ENTRYPOINT ["/usr/local/bin/dumb-init", "/usr/local/bin/consul-dataplane"]
 
+# FIPS release-default release image
+# -----------------------------------
+FROM gcr.io/distroless/base-debian11 AS release-fips-default
```
curtbushko, line 58: Any specific reason to use debian here?  I think all of our container images are based on alpine which makes it easier to track and fix CVEs.

skpratt: This is an exact copy of the release-default target above, so if we want to change it there, we could then update this once that has been done

curtbushko: Ahhh, I missed that. It should be ok!

curtbushko (approved):

PR #164, opened 2023-06-20
Added `config-file` flag to accept file based input params

Ganeshrockz:
**Why is the change required?**

As part of the work needed to migrate `consul-ecs` to use `consul-dataplane` and remove consul clients from the ECS tasks, we would need to make the consul ecs control plane configure the dataplane config. Since ECS tasks are immutable by nature, we can't call the dataplane CLI command with the appropriate parameters similar to how it is being done in consul-k8s.

According to the RFC, it was decided that the Consul ECS control plane would take of care of writing the configuration needed to start dataplane to a shared volume and pass the filename as an input parameter to the `consul-dataplane` command. Dataplane would then configure itself based on the config values present in the file.

**What has changed?**

This PR introduces an additional string input parameter `-config-file` which accepts a `json` file for now. We can later extend it to accepting `hcl` files if needed. 

**How is the change implemented?**

CLI flag inputs (if provided) still holds the highest precedence even if the same config input field is present the config file. Upon startup, dataplane prepares a default config based on the default values present already for various fields. Following that, dataplane reads the configuration present in the `json` file, unmarshals the input and prepares a config struct based on the flags present in the file. It then merges the second config onto the first config if the values present in the second config aren't the default ones. Finally dataplane constructs a third struct based on the CLI flag inputs and merges it to the previously formed config. In this way, we can make sure that the right order of precedence is maintained when accepting and processing inputs

**How was the change tested?**

- Unit tests (covering various input generation flows
- CI


cmd/consul-dataplane/config.go
```diff
@@ -0,0 +1,294 @@
...
+		},
+		Service: &consuldp.ServiceConfig{
+			NodeName:  f.nodeName,
+			NodeID:    f.nodeID,
+			ServiceID: f.serviceID,
+			Namespace: f.namespace,
+			Partition: f.partition,
+		},
+		Logging: &consuldp.LoggingConfig{
+			Name:     "consul-dataplane",
```
absolutelightning, line 197: This string "consul-dataplane" is at 5 places, we should move this to a constant.

cmd/consul-dataplane/merge.go
```diff
@@ -0,0 +1,269 @@
...
+
+	if c2.Addresses != "" {
+		c1.Addresses = c2.Addresses
+	}
+
+	if c2.GRPCPort != int(0) && c2.GRPCPort != DefaultGRPCPort {
+		c1.GRPCPort = c2.GRPCPort
+	}
+
+	if c2.ServerWatchDisabled {
```
absolutelightning, line 38: lets say in config file `ServerWatchDisabled` was true and in command line flag it was false. This code will not override the flag.

cmd/consul-dataplane/merge.go
```diff
@@ -0,0 +1,269 @@
...
+		}
+
+		if c2.Credentials.Static.Token != "" {
+			c1.Credentials.Static.Token = c2.Credentials.Static.Token
+		}
+	}
+
+	if c2.TLS != nil {
+		if c2.TLS.Disabled {
+			c1.TLS.Disabled = true
```
absolutelightning, line 82: I think - if `TLS.Disabled` is true in config file but false in flag, ideally final config should be false right?.
This code will not set it to false.

cmd/consul-dataplane/merge.go
Ganeshrockz: I am aware of this case and looking for ways to solve it. It stems from the fact that the flag inputs assume a default value when it is not given via the inputs by the user. Let's say we have the following config

```json
"telemetry": {
   "useCentralConfig": false
}
```

and the user gives the same input via the flag

```
consul-dataplane --use-central-config true --addr abc.dc1
```

In this case, we would expect the `central-config` to be true but it becomes false per the current design. Now let's say the user provided the following input

```
consul-dataplane --addr abc.dc1
```

In this scenario, `central-config` assumes a default value of `true` because the user hasn't provided it via the inputs but the end result has to be `false` because we should respect the value given via the config file. There isn't a clean way to differentiate if a value for a flag is given by the user (user gives the default value) or is the default value added by the system.

absolutelightning: Will it be better to we can make command line flag mandatory and not assume default values?
That way we always override the value in config file to command line value.

absolutelightning: also if somehow we differentiate that is it default value or given by user. Then it will solve our problem.
otherwise another solution is we can keep a number instead of boolean 
0 = default
1 = true
2 = false
just sharing possibility, but needs dicussion.

Ganeshrockz: > Will it be better to we can make command line flag mandatory and not assume default values?

This will be a breaking change with the current way clients are using dataplane. We should probably look into this.

> otherwise another solution is we can keep a number instead of boolean

There are also string based input fields. So it doesn't seem to be trivial.

One simple way to approach this change is to allow either passing values by flag or by config file. Since dataplane is an internal component we could start with that model for simplicity. I want to hear from reviewers as well around this.

cmd/consul-dataplane/merge.go
```diff
@@ -0,0 +1,269 @@
...
+// 2. If an attribute of c2 is a slice/map, then it overrides the same field in c1
+func mergeConfigs(c1, c2 *consuldp.Config) {
+	mergeConsulConfigs(c1.Consul, c2.Consul)
+	mergeServiceConfigs(c1.Service, c2.Service)
+	mergeTelemetryConfigs(c1.Telemetry, c2.Telemetry)
+	mergeLoggingConfigs(c1.Logging, c2.Logging)
+	mergeEnvoyConfigs(c1.Envoy, c2.Envoy)
+	mergeXDSServerConfigs(c1.XDSServer, c2.XDSServer)
+	mergeDNSServerConfigs(c1.DNSServer, c2.DNSServer)
+}
```
absolutelightning, line 23: @Ganeshrockz - can we please consider - https://github.com/imdario/mergo, to make code more beautiful.
That library uses reflection to figure out keys dynamically. I think that will be a better approach instead of setting each key value pair.

cmd/consul-dataplane/merge.go
Ganeshrockz: This should be fixed. I modified the whole approach to make stuff work.

cmd/consul-dataplane/flags.go
```diff
@@ -57,3 +58,170 @@ func MapVar(v flag.Value, name, env, usage string) {
 func includeEnvUsage(env, usage string) string {
 	return fmt.Sprintf("%s Environment variable: %s.", usage, env)
 }
+
+// stringPtrValue is a flag.Value which stores the value in a *string.
+// If the value was not set the pointer is nil.
+type stringPtrValue struct {
+	v **string
```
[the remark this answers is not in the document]
Ganeshrockz: I thought of going with zero values for struct fields, but it becomes very tricky to override the flags when they are given via the file and CLI.

https://github.com/hashicorp/consul-dataplane/pull/164#discussion_r1234750138 explains the corner case and the issue with relying on zero valued fields.

> The pattern is a little unusual to my eyes

I took inspiration from the consul core codebase which seems to be applying configs in a similar manner. They start with a pointer based fields in a struct and finally construct the runtime config needed for the agents.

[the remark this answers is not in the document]
pglass: Ah, this makes sense. Thanks for the explanation!

cmd/consul-dataplane/config_test.go
```diff
@@ -0,0 +1,547 @@
...
+		wantErr         bool
+	}
+
+	testCases := []testCase{
+		{
+			desc: "able to generate config properly when the config file input is empty",
+			flagOpts: func() (*FlagOpts, error) {
+				return generateFlagOpts()
+			},
+			assertConfig: func(cfg *consuldp.Config, flagOpts *FlagOpts) bool {
```
pglass, line 32: nit/suggestion:  I don't think the `cfg` parameter is used to construct the `expectedCfg` in any of these `assertConfig` functions, so it doesn't seem necessary to pass it in.

Instead of having this function execute the assertion, it could return the expected config:

```suggestion
			makeExpectedConfig: func(flagOpts *FlagOpts) *consuldp.Config { ... }
```

Then in the test body down below (on line 455), we can replace `require.True(t, tc.assertConfig(cfg, opts)` with `require.Equal(t, tc.makeExpectedConfig(opts), cfg)` (which does a deep equality check).

What do you think?

cmd/consul-dataplane/config_test.go
```diff
@@ -0,0 +1,547 @@
...
+					"logging": {
+					  "logLevel": "info",
+					  "logJSON": false
+					}
+				  }`
+
+				err := os.WriteFile("test.json", []byte(inputJson), 0600)
+				if err != nil {
+					return err
+				}
```
pglass, line 244: minor: Have this function accept a `*testing.T` so that we can run assertions here and use `t.Cleanup` to run cleanups after the test finishes.

```go
			writeConfigFile: func(t *testing.T) {
```

And then this becomes:

```suggestion
				err := os.WriteFile("test.json", []byte(inputJson), 0600)
				require.NoError(t, err)
				t.Cleanup(func() { _ = os.Remove("test.json") }
```

I prefer this because the cleanups are added right next to where the file is written. (And can make it easier to refactor, if we want to move this function somewhere else because the cleanups are self-contained). What do you think?

cmd/consul-dataplane/merge.go
Ganeshrockz: Thanks for the suggestion. I was looking for alternatives to the current approach and mergo seems to be the perfect candidate. Added the same.

@pglass It did not increase the image size a lot.

pglass (approved):
Few other minor comments, but this looks good to me

cmd/consul-dataplane/config_test.go
Ganeshrockz: Made the change 👍

cmd/consul-dataplane/config.go
```diff
@@ -0,0 +1,315 @@
...
+	if err != nil {
+		return nil, err
+	}
+
+	return cfg, nil
+}
+
+// Constructs a config with the default values
+func buildDefaultConsulDPFlags() (*DataplaneConfigFlags, error) {
+	data := `
```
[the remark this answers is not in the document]
Ganeshrockz: Since the fields of `DataplaneConfigFlags` are mostly pointers, it kind of looks ugly referencing the const variables inside the struct. We can't define `const` variables for these default flags and assign them as `&DefineConsulFlag` to those fields. We would have to fallback to primitive types for these variables which doesn't look that good as well. So I went ahead with a hardcoded json that is easier for people to make changes. I can reiterate on this if people see further concerns.

absolutelightning (approved):

PR #165, opened 2023-06-20
Fix broken proxy tests in release/1.2.x

curtbushko:
We broke the proxy tests when we reverted the Proxy Lifecycle changes.


DanStough (approved):
Thank you!

nathancoleman (approved):

PR #172, opened 2023-06-22
Send the Namespace/Partition as gRPC metadata headers in the request to the Consul DNS service.

mkeeler:
## Description

By passing along the tenant of the service being proxied we can more intelligently select the correct partition/namespace to use as defaults for DNS.

This also passes along the Consul ACL token as the x-consul-token grpc metadata. Eventually Consul should be able to use this token instead of the servers default token for authorizing access.

## TODO

- [ ] Tests
- [ ] Support for handling this metadata in Consul (other PRs are in the works for that)
- [ ] A review of whether this was the best way to get the tenancy information to the DNS server

### DIFF
--- a/pkg/consuldp/bootstrap.go
+++ b/pkg/consuldp/bootstrap.go
@@ -55,10 +55,9 @@ func (cdp *ConsulDataplane) bootstrapConfig(ctx context.Context) (*bootstrap.Boo
 	}
 
 	// store the final resolved service for others to use.
-	cdp.resolvedService = ServiceConfig{
+	cdp.resolvedProxyConfig = ProxyConfig{
 		NodeName:  rsp.NodeName,
-		NodeID:    rsp.NodeId,
-		ServiceID: cdp.cfg.Service.ServiceID,
+		ProxyID:   cdp.cfg.Proxy.ProxyID,
 		Namespace: rsp.Namespace,
 		Partition: rsp.Partition,
 	}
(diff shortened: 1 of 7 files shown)


pkg/consuldp/bootstrap.go
```diff
@@ -54,6 +54,14 @@ func (cdp *ConsulDataplane) bootstrapConfig(ctx context.Context) (*bootstrap.Boo
 		return nil, nil, fmt.Errorf("failed to get envoy bootstrap params: %w", err)
 	}
 
+	// store the final resolved service for others to use.
+	cdp.resolvedProxyConfig = ProxyConfig{
+		NodeName:  rsp.NodeName,
+		ProxyID:   cdp.cfg.Proxy.ProxyID,
+		Namespace: rsp.Namespace,
+		Partition: rsp.Partition,
+	}
```
nathancoleman, line 63: Would you expect that the node, namespace + partition would sometimes differ between `rsp` and `cdp.cfg.Proxy`? I ask because I was curious why we can't just use `cfg.cfg.Proxy` directly vs. hanging an additional `ProxyConfig` on the struct

DanStough: Matt might know differently, but I assumed it was that the serviceaccount/Token used to make the bootstrapping could imply a namespace/partition that wasn't specified using the config flags.

jmurret (approved):
Looks good.  👏

nathancoleman (approved):

PR #188, opened 2023-06-28
Net 4451/windows dataplane fix

aahel:
this pr is based on the changes from this [pr](https://github.com/hashicorp/consul-dataplane/pull/46).

- Switch bootstrap config file from a named Unix pipe to a tempfile, since there shouldn't be anything sensitive in the bootstrap config file
- Remove x-consul-token from the bootstrap config template since the token is injected into the gRPC stream by an interceptor

Testing method :- 
Have tested this building building a new windows dataplane image and deploying consul along with dashboard and counting service in a gke windows cluster.


internal/bootstrap/bootstrap_config.go
```diff
@@ -846,7 +846,8 @@ func appendTelemetryCollectorConfig(args *BootstrapTplArgs, telemetryCollectorBi
 			"envoy_grpc": {
 			  "cluster_name": "consul_telemetry_collector_loopback"
 			}
-		  }
+		  },
+		  "emit_tags_as_labels": true
```
pglass, line 850: There's another PR up which adds these: https://github.com/hashicorp/consul-dataplane/pull/184. I'd recommend we merge that one first so that it has the right changelog entry (I approved) and then rebase this PR

pkg/envoy/proxy.go
```diff
@@ -343,40 +344,17 @@ func writeBootstrapConfig(cfg []byte) (string, func() error, error) {
...
-	}
-
-	// O_WRONLY causes OpenFile to block until there's a reader (Envoy). Opening
-	// the pipe with O_RDWR wouldn't block but would result in just sending stuff
-	// to ourself.
-	//
-	// TODO(boxofrad): We don't have a way to cancel this goroutine. If the Envoy
-	// process never opens the other end of the pipe this will hang forever. The
-	// workaround we use in `consul connect envoy` is to write to the pipe in a
-	// subprocess that self-destructs after 10 minutes.
```
pglass, line 357: We should add a changelog entry for this in `.changelog/`

internal/bootstrap/bootstrap_config.go
aahel: sure. will wait for it be merged.

jjti: I just merged https://github.com/hashicorp/consul-dataplane/pull/184, thanks for waiting ya'll

pkg/envoy/proxy.go
aahel: done

.changelog/188.txt
```diff
@@ -0,0 +1,3 @@
+```release-note:bug
+ fixed the flow of passing bootstrap config to envoy by writing it to a file instead of using syscall.Mkfifo which is unavailable on Windows.
```
pglass, line 2: I think this is an improvement (or possibly feature). We are adding support for a new platform (basically, it was not a bug that it didn't work on Windows because we had not intended for it to work on Windows originally).

What do you think?

`````suggestion
```release-note:improvement
In order to support Windows, write Envoy bootstrap configuration to a regular file instead of a named pipe.
`````

**Review by pglass — APPROVED** [review]
Approved. Feel free to address the changelog comment how you like.

File: pkg/envoy/get_process_attr_windows.go [hunk 4]
```diff
@@ -0,0 +1,12 @@
...
+
+package envoy
+
+import "syscall"
+
+func getProcessAttr() *syscall.SysProcAttr {
+	return &syscall.SysProcAttr{
+		CreationFlags: syscall.CREATE_NEW_PROCESS_GROUP,
+	}
+}
```
aahel, line 12: @pglass  i had to create a seperate windows and linux versions of this syscall to create process group. should i mention this in changelog ?

pkg/envoy/get_process_attr.go
```diff
@@ -0,0 +1,12 @@
...
+
+package envoy
+
+import "syscall"
+
+func getProcessAttr() *syscall.SysProcAttr {
+	return &syscall.SysProcAttr{
+		Setpgid: true,
+	}
+}
```
aahel, line 12: this is the non windows one

pkg/envoy/get_process_attr_windows.go
pglass: I don't think it's necessary to call this out in the changelog.

PR #216, opened 2023-07-19
Differentiate linux package names

skpratt:
This avoids a name collision in linux package managers, allowing people to get non-FIPSs with i.e. "apt-get install consul-dataplane" and FIPS with "apt-get install consul-dataplane-fips"

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -74,8 +74,8 @@ jobs:
           - {goos: "linux", goarch: "arm64"}
           - {goos: "linux", goarch: "386"}
           - {goos: "linux", goarch: "amd64"}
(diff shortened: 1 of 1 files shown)


zalimeni (approved):
LGTM.

@skpratt are you planning to update the release docs for each of the repos getting this change? I think we'll want to call this out in the package validation steps; AFAIK we haven't fully automated that yet across repos.

skpratt:
> LGTM.
> 
> @skpratt are you planning to update the release docs for each of the repos getting this change? I think we'll want to call this out in the package validation steps; AFAIK we haven't fully automated that yet across repos.

Yes, I'll have releases docs PRs up later today

PR #239, opened 2023-08-11
net 2898, graceful_startup

trevorLeonHC:
Adding graceful_startup endpoint and function that will be called by dataplane postStart hook. Graceful_startup will block all other containers from starting up until envoy dataplane is determined to be ready.


t-eckert:
You have a few unit tests to clean up. Because the `/graceful_startup` is now set as a default configuration. The configuration tests that are looking at the default values are failing because they expect `""` and are getting `"/graceful_startup"`.

cmd/consul-dataplane/main.go
```diff
@@ -102,7 +102,9 @@ func init() {
 
 	// Default is false, may be useful for debugging unexpected termination.
 	BoolVar(flags, &flagOpts.dataplaneConfig.Envoy.DumpEnvoyConfigOnExitEnabled, "dump-envoy-config-on-exit", "DP_DUMP_ENVOY_CONFIG_ON_EXIT", "Call the Envoy /config_dump endpoint during consul-dataplane controlled shutdown.")
-
```
curtbushko, line 105: **issue:** Please move this up a few lines so that it is with the other graceful flags.  Also, please remove `//########`

cmd/consul-dataplane/main.go
```diff
@@ -102,7 +102,9 @@ func init() {
...
+	//#########################
```
curtbushko, line 105: **issue:** Please move this up a few lines so that it is with the other graceful flags. Also, please remove //########

pkg/consuldp/config.go
```diff
@@ -297,6 +297,11 @@ type EnvoyConfig struct {
 	GracefulPort int
 	// DumpEnvoyConfigOnExitEnabled configures whether to call Envoy's /config_dump endpoint during consul-dataplane controlled shutdown.
 	DumpEnvoyConfigOnExitEnabled bool
+	//##############
```
curtbushko, line 300: **issue:** Same as above. Please move this up a few lines so that it is with the other graceful flags. Also, please remove //########

pkg/consuldp/lifecycle.go
```diff
@@ -211,3 +216,60 @@ func (m *lifecycleConfig) gracefulShutdown() {
...
+	//in order to block application container.
+	rw.WriteHeader(http.StatusOK)
+
+}
+
+// gracefulStartup blocks until the startup grace period has elapsed or we have confirmed that
+// Envoy proxy is ready.
+func (m *lifecycleConfig) gracefulStartup() {
+	m.logger.Info("Blocking container startup until Envoy ready / grace period elapsed")
+	timeout := time.Duration(m.shutdownGracePeriodSeconds) * time.Second
```
curtbushko, line 233: **issue:** m.shutdownGracePeriodSeconds -> m.startupGracePeriodSeconds

pkg/consuldp/lifecycle.go
```diff
@@ -211,3 +216,60 @@ func (m *lifecycleConfig) gracefulShutdown() {
...
+	ctx, cancel := context.WithTimeout(context.Background(), timeout)
```
curtbushko, line 234: **issue:** I am not 100% on this `loop:`. I think you may be able to send a message to ctx.Done() instead for both the timeout and ready aspect

curtbushko (changes requested):
Thanks! This is looking pretty great!  I have a few requests but I am not too sure about the loop thing.

cmd/consul-dataplane/config.go
```diff
@@ -121,6 +121,9 @@ type EnvoyFlags struct {
 	GracefulShutdownPath          *string `json:"gracefulShutdownPath,omitempty"`
 	GracefulPort                  *int    `json:"gracefulPort,omitempty"`
 	DumpEnvoyConfigOnExitEnabled  *bool   `json:"dumpEnvoyConfigOnExitEnabled,omitempty"`
+
+	StartupGracePeriodSeconds *int    `json:"startupGracePeriodSeconds,omitempty"`
+	GracefulStartupPath       *string `json:"gracefulStartupPath,omitempty"`
```
t-eckert, line 126: I know the other fields in this struct don't have any comments, but I think it would be good to start a trend and comment your fields here.

pkg/consuldp/lifecycle.go
```diff
@@ -211,3 +216,60 @@ func (m *lifecycleConfig) gracefulShutdown() {
...
+
+			}
+		}
+
+	}()
+
+	wg.Wait()
+
+	if !envoyReady {
+		m.logger.Info("startup grace period reached before envoy ready")
```
DanStough, line 272: Is there a `warn` level? Feels like we would want to surface this to the operator.

pkg/consuldp/lifecycle.go
```diff
@@ -211,3 +216,60 @@ func (m *lifecycleConfig) gracefulShutdown() {
...
+	//in order to block application container.
+	m.gracefulStartup()
+	rw.WriteHeader(http.StatusOK)
+
+}
+
+// gracefulStartup blocks until the startup grace period has elapsed or we have confirmed that
+// Envoy proxy is ready.
+func (m *lifecycleConfig) gracefulStartup() {
+	timeout := time.Duration(m.startupGracePeriodSeconds) * time.Second
```
DanStough, line 232: Maybe bypass all this if the grace period is unset/0? That will reduce some of the log chatter.

.DS_Store
t-eckert, line 1: Be sure to delete this file. It's binary unique to Mac OS that defines the directory metadata for the Finder. 

You'll encounter this a lot when developing on a Mac using Git so I have it added to my global `.gitignore`.

The command to add `.DS_Store` to your global `.gitignore` is:

```shell
git config --global core.excludesfile "~/.gitignore" &&  echo .DS_Store >> ~/.gitignore
```

pkg/consuldp/config.go
```diff
@@ -293,6 +293,10 @@ type EnvoyConfig struct {
 	ShutdownGracePeriodSeconds int
 	// GracefulShutdownPath is the path on which the HTTP endpoint to initiate a graceful shutdown of Envoy is served
 	GracefulShutdownPath string
+	//StartupGracePeriodSeconds is the amount of time to block application after startup for Envoy proxy to be ready.
+	StartupGracePeriodSeconds int
+	// GracefulStartupPath is the path on which the HTTP endpoint to initiate a graceful startup of Envoy is served
```
t-eckert, line 298:
```suggestion
	// GracefulStartupPath is the path where the HTTP endpoint to initiate a graceful startup of Envoy is served
```

pkg/consuldp/lifecycle.go
```diff
@@ -211,3 +207,81 @@ func (m *lifecycleConfig) gracefulShutdown() {
 	// Wait for context timeout to elapse
 	wg.Wait()
 }
+
+func (m *lifecycleConfig) gracefulStartupHandler(rw http.ResponseWriter, _ *http.Request) {
+
+	//Unlike in gracefulShutdown, we want to delay the OK response until envoy is ready
+	//in order to block application container.
```
t-eckert, line 214: This would be a great place to emit a log. "Request received to graceful startup. Returning when Dataplane is in a ready state."

pkg/consuldp/lifecycle_test.go
```diff
@@ -38,35 +40,35 @@ func TestLifecycleServerClosed(t *testing.T) {
 	require.Eventually(t, func() bool {
 		return !m.running
 	}, time.Second*2, time.Second)
-
 }
 
-func TestLifecycleServerEnabled(t *testing.T) {
+func TestLifecycleServer(t *testing.T) {
```
t-eckert, line 45: Let's add a comment here on what this test tests.

pkg/consuldp/lifecycle.go
```diff
@@ -211,3 +207,81 @@ func (m *lifecycleConfig) gracefulShutdown() {
...
+	defer cancel()
+
+	envoyReady := false
+	var wg sync.WaitGroup
+	wg.Add(1)
+	envoyStatus := make(chan bool)
+
+	go func() {
+		defer wg.Done()
+		envoyReady, _ = m.proxy.Ready()
```
t-eckert, line 238: Both here and on line 253, let's log out the error if it is returned.

t-eckert (changes requested):
Added a few small comments, but overall this is looking great!

pkg/consuldp/lifecycle_test.go
curtbushko, line 47: **suggestion:** Even though there is some duplication in setup code, you may want to split this into two tests as you are doing startup specific things even when doing shutdown tests. I think it is making it more complicated than it needs to be.

pkg/consuldp/lifecycle.go
curtbushko: @trevorLeonHC I talk to my TL and this is what he suggested as maybe a better way to avoid the channel complications...

Something to consider...
```
func isReady() bool {
	time.Sleep(1 * time.Second)
	return true
}

func main() {
	ctx, cancel := context.WithTimeout(context.Background(), 3*time.Second)
	defer cancel()

	var ready atomic.Bool
	go func() {
		for ctx.Err() == nil {
			r := isReady()
			if r {
				ready.Store(true)
				cancel()
				break
			}
			// maybe put in a sleep of 50ms or something to prevent eating CPU if isReady could return quickly.
		}
	}()

	<-ctx.Done()
	fmt.Println("Got value", ready.Load())
}
```

pkg/consuldp/config.go
```diff
@@ -293,6 +293,10 @@ type EnvoyConfig struct {
 	ShutdownGracePeriodSeconds int
 	// GracefulShutdownPath is the path on which the HTTP endpoint to initiate a graceful shutdown of Envoy is served
 	GracefulShutdownPath string
+	//StartupGracePeriodSeconds is the amount of time to block application after startup for Envoy proxy to be ready.
+	StartupGracePeriodSeconds int
+	// GracefulStartupPath is the path where the HTTP endpoint to initiate a graceful startup of Envoy is served
```
t-eckert, line 298:
```suggestion
	// StartupGracePeriodSeconds is the amount of time to block application after startup for Envoy proxy to be ready.
	StartupGracePeriodSeconds int
	// GracefulStartupPath is the path where the HTTP endpoint to initiate a graceful startup of Envoy is served.
```

pkg/consuldp/lifecycle.go
```diff
@@ -211,3 +208,62 @@ func (m *lifecycleConfig) gracefulShutdown() {
...
 }
+
+func (m *lifecycleConfig) gracefulStartupHandler(rw http.ResponseWriter, _ *http.Request) {
+
+	//Unlike in gracefulShutdown, we want to delay the OK response until envoy is ready
+	//in order to block application container.
+	m.gracefulStartup()
+	rw.WriteHeader(http.StatusOK)
+
+}
```
t-eckert, line 219:
```suggestion
func (m *lifecycleConfig) gracefulStartupHandler(rw http.ResponseWriter, _ *http.Request) {
	//Unlike in gracefulShutdown, we want to delay the OK response until envoy is ready
	//in order to block application container.
	m.gracefulStartup()
	rw.WriteHeader(http.StatusOK)
}
```

pkg/consuldp/lifecycle_test.go
```diff
@@ -38,35 +40,153 @@ func TestLifecycleServerClosed(t *testing.T) {
 	require.Eventually(t, func() bool {
 		return !m.running
 	}, time.Second*2, time.Second)
+}
+func TestLifecycleServer_Startup(t *testing.T) {
```
t-eckert, line 44: Let's add a blank line here and a comment to describe what the test is testing.

pkg/consuldp/lifecycle_test.go
```diff
@@ -21,6 +21,8 @@ var (
 	envoyAdminAddr = "127.0.0.1"
 )
 
+// TestLifecycleServerClosed tests that the lifecycle manager properly starts up
+// and shuts down when the context passed into it is cancelled.
```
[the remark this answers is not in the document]
t-eckert: Nevermind, I misspoke.

pkg/consuldp/lifecycle_test.go
```diff
@@ -38,35 +40,153 @@ func TestLifecycleServerClosed(t *testing.T) {
...
+			require.NoError(t, err)
+			require.NotNil(t, body)
+		})
+	}
 
 }
 
-func TestLifecycleServerEnabled(t *testing.T) {
+// TestLifecycleServer tests the different functions of the lifecycle server,
+// using a mock proxy, including graceful_shutdown and graceful_startup.
```
t-eckert, line 165: It's _this_ test's comment we want to edit.

pkg/consuldp/lifecycle.go
t-eckert: This was an awesome solution! I learned from it. Implemented!

t-eckert (approved):
Left a few non-blocking comments. Great work on getting code into production!

zalimeni:
Removing backport labels for clarity as this feature was not backported (exists in 1.3.x+)

PR #242, opened 2023-08-17
Make consul dataplane handle bootstrap param response for V2 resources

ishustava:
* Handle bootstrap_config param
* Use new request and response params
* Make flags/config more generic so that they can accomodate services or workloads and deprecate old flags/config.
Part of NET-5221

### DIFF
--- a/cmd/consul-dataplane/config.go
+++ b/cmd/consul-dataplane/config.go
@@ -9,6 +9,7 @@ import (
 	"strings"
 
 	"dario.cat/mergo"
+
 	"github.com/hashicorp/consul-dataplane/pkg/consuldp"
 )
 
@@ -77,13 +78,13 @@ type ServiceFlags struct {
 	Partition     *string `json:"partition,omitempty"`
 }
 
-func (sf ServiceFlags) IsEmpty() bool {
-	return sf.NodeName == nil &&
-		sf.NodeID == nil &&
-		sf.ServiceID == nil &&
-		sf.ServiceIDPath == nil &&
-		sf.Namespace == nil &&
-		sf.Partition == nil
+func (pf ProxyFlags) IsEmpty() bool {
+	return pf.NodeName == nil &&
+		pf.NodeID == nil &&
+		pf.ID == nil &&
+		pf.IDPath == nil &&
+		pf.Namespace == nil &&
+		pf.Partition == nil
 }
 
 type ProxyFlags struct {
@@ -268,21 +269,21 @@ func buildDefaultConsulDPFlags() (DataplaneConfigFlags, error) {
 func constructRuntimeConfig(cfg DataplaneConfigFlags, extraArgs []string) (*consuldp.Config, error) {
 	// Handle deprecated service flags.
 	var proxyCfg consuldp.ProxyConfig
-	if !cfg.Service.IsEmpty() {
-		proxyCfg = consuldp.ProxyConfig{
-			NodeName:  stringVal(cfg.Service.NodeName),
-			NodeID:    stringVal(cfg.Service.NodeID),
-			ProxyID:   stringVal(cfg.Service.ServiceID),
-			Namespace: stringVal(cfg.Service.Namespace),
-			Partition: stringVal(cfg.Service.Partition),
-		}
-	} else {
+	if !cfg.Proxy.IsEmpty() {
 		proxyCfg = consuldp.ProxyConfig{
 			NodeName:  stringVal(cfg.Proxy.NodeName),
 			NodeID:    stringVal(cfg.Proxy.NodeID),
 			ProxyID:   stringVal(cfg.Proxy.ID),
 			Namespace: stringVal(cfg.Proxy.Namespace),
 			Partition: stringVal(cfg.Proxy.Partition),
+		}
+	} else {
+		proxyCfg = consuldp.ProxyConfig{
+			NodeName:  stringVal(cfg.Service.NodeName),
+			NodeID:    stringVal(cfg.Service.NodeID),
+			ProxyID:   stringVal(cfg.Service.ServiceID),
+			Namespace: stringVal(cfg.Service.Namespace),
+			Partition: stringVal(cfg.Service.Partition),
 		}
 	}
 
--- a/cmd/consul-dataplane/config.go
+++ b/cmd/consul-dataplane/config.go
@@ -22,6 +22,7 @@ type FlagOpts struct {
 type DataplaneConfigFlags struct {
 	Consul    ConsulFlags    `json:"consul,omitempty"`
 	Service   ServiceFlags   `json:"service,omitempty"`
+	Proxy     ProxyFlags     `json:"proxy,omitempty"`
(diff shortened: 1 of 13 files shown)


pkg/consuldp/consul_dataplane.go
```diff
@@ -77,11 +77,11 @@ func validateConfig(cfg *Config) error {
 		return errors.New("consul addresses not specified")
 	case cfg.Consul.GRPCPort == 0:
 		return errors.New("consul server gRPC port not specified")
-	case cfg.Service == nil:
+	case cfg.Proxy == nil:
 		return errors.New("service details not specified")
```
DanStough, line 81:
```suggestion
		return errors.New("proxy details not specified")
```

pkg/consuldp/consul_dataplane.go
```diff
@@ -77,11 +77,11 @@ func validateConfig(cfg *Config) error {
...
-	case cfg.Service.NodeID == "" && cfg.Service.NodeName == "":
+	case cfg.Proxy.NodeID == "" && cfg.Proxy.NodeName == "":
 		return errors.New("node name or ID not specified")
-	case cfg.Service.ServiceID == "":
+	case cfg.Proxy.ProxyID == "":
 		return errors.New("proxy service ID not specified")
```
DanStough, line 85:
```suggestion
		return errors.New("proxy ID not specified")
```

go.mod
```diff
@@ -7,7 +7,7 @@ require (
 	github.com/adamthesax/grpc-proxy v0.0.0-20220525203857-13e92d14f87a
 	github.com/armon/go-metrics v0.4.1
 	github.com/hashicorp/consul-server-connection-manager v0.1.3
-	github.com/hashicorp/consul/proto-public v0.4.0
+	github.com/hashicorp/consul/proto-public v0.1.2-0.20230817221841-aac2dac759c6
```
nathancoleman, line 10: Is this intentional?

cmd/consul-dataplane/config_test.go
```diff
@@ -103,9 +103,189 @@ func TestConfigGeneration(t *testing.T) {
...
+							CertFile:      "prom-ca-cert.pem",
+							KeyFile:       "prom-key.pem",
+						},
+					},
+				}
+			},
+			wantErr: false,
+		},
+		{
+			desc: "able to override all the config fields with CLI flags when using service flags",
```
DanStough, line 183: IMHO this test case is less interesting to duplicate because none of the flags overlap with service vs proxy config. However, `"test whether CLI flag values override the file values` does have a "service" config block, so I do think there should be a test there.

pkg/consuldp/bootstrap.go
```diff
@@ -76,6 +77,11 @@ func (cdp *ConsulDataplane) bootstrapConfig(ctx context.Context) (*bootstrap.Boo
 		PrometheusScrapePath:  prom.ScrapePath,
 	}
 
+	if rsp.ClusterName != "" {
```
DanStough, line 80: Same comment about naming from the Consul PR if this is supposed to be `identity`.

DanStough (commented):
See the question about whether or not the tenancy information is required now.

This is probably a future concern, but I imagined in a world or migration where V1 and V2 run in parallel, dataplane would need some way to indicate it wants a V1 or V2 style response if it's using the same, backwards-compatible endpoint. Is that something we should plan for?

go.mod
DanStough: I think the answer is yes because she is pulling in an unreleased version of this module with all the new V2 APIs. We do something similar in `consul-k8s` with new versions of the `api` module until it is released. Go does a weird thing where it uses the last tag that exists on the `main` branch along with part of the SHA to identify the commit to pull in.

cmd/consul-dataplane/config.go
```diff
@@ -76,6 +77,24 @@ type ServiceFlags struct {
...
+		sf.NodeID == nil &&
+		sf.ServiceID == nil &&
+		sf.ServiceIDPath == nil &&
+		sf.Namespace == nil &&
+		sf.Partition == nil
+}
+
+type ProxyFlags struct {
+	NodeName  *string `json:"nodeName,omitempty"`
+	NodeID    *string `json:"nodeID,omitempty"`
```
mkeeler, line 91: Are these needed with v2. All workload ids must be unique within a tenant so it seems like we may not need these any longer.

cmd/consul-dataplane/main.go
```diff
@@ -43,12 +43,35 @@ func init() {
...
+		"The ID of the Consul node to which the proxy service instance is registered."+
+			"In Consul's V2 Catalog API, this value is ignored.")
+	StringVar(flags, &flagOpts.dataplaneConfig.Proxy.ID, "proxy-id", "DP_PROXY_ID",
+		"In Consul's V1 Catalog API, the proxy service instance's ID."+
+			"In Consul's V2 Catalog API, the workload ID associated with the proxy.")
+	StringVar(flags, &flagOpts.dataplaneConfig.Proxy.IDPath, "proxy-id-path", "DP_PROXY_ID_PATH",
+		"In Consul's V1 Catalog API, the path to a file containing the proxy service instance's ID."+
+			"In Consul's V2 Catalog API, the path to a file containing the workload ID associated with the proxy.")
+	StringVar(flags, &flagOpts.dataplaneConfig.Proxy.Namespace, "proxy-namespace", "DP_PROXY_NAMESPACE",
+		"The Consul Enterprise namespace in which the proxy service instance (V1 API) or workload (V2 API) is registered.")
```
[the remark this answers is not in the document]
mkeeler: Consul will have to fill in defaults for you. For v2 things, that hasn't been implemented yet but it will have to be there pre-GA.

cmd/consul-dataplane/config.go
ishustava: These are not needed with V2, however, these flags can be used for both v1 and v2. My intention is to deprecate old flags and to have one set of generically named flags that could be used against either API.

cmd/consul-dataplane/config_test.go
ishustava: I think this test is still valuable. I believe it's testing that every possible config value can be set via CLI flags. In this case, we need to still be able to check that if I provide proxy flags, those end up in dataplane config (even without overlapping service flags). We could create a test where we're only testing that subset of flags but I thought given that service flags are deprecated, it'd be easier to remove in the future because we can just keep the other test case and remove this one without sacrificing any test coverage.

>"test whether CLI flag values override the file values does have a "service" config block, so I do think there should be a test there.

I've added two more cases. One to test that cli flags override file values when using proxy flags and another one to test that proxy flags are preferred over service flags.

pkg/consuldp/bootstrap.go
ishustava: Changed it to identity.

DanStough (approved):
🎉

DanStough (approved):
2 x 🎉

PR #253, opened 2023-08-30
NET-5186 Fix issue where TMPDIR envar is not visible from consul-dataplane binary

nathancoleman:


Dockerfile
```diff
@@ -61,8 +71,9 @@ LABEL name=${BIN_NAME}\
 
 COPY --from=dumb-init /usr/bin/dumb-init /usr/local/bin/
 COPY --from=go-discover /go/bin/discover /usr/local/bin/
-COPY --from=setcap /usr/local/bin/envoy /usr/local/bin/
-COPY --from=setcap /usr/local/bin/$BIN_NAME /usr/local/bin/
+COPY --from=setcap-envoy-binary /usr/local/bin/envoy /usr/local/bin/
+COPY dist/$TARGETOS/$TARGETARCH/$BIN_NAME /usr/local/bin/
```
nathancoleman, line 75: All release images now pull in the consul-dataplane binary directly -- it no longer has `setcap` run on it

Dockerfile
```diff
@@ -90,9 +101,10 @@ LABEL name=${BIN_NAME}\
       description="Consul dataplane manages the proxy that runs within the data plane layer of Consul Service Mesh."
 
 COPY --from=go-discover /go/bin/discover /usr/local/bin/
-COPY --from=envoy-fips-binary /usr/local/bin/envoy /usr/local/bin/
+COPY --from=setcap-envoy-fips-binary /usr/local/bin/envoy /usr/local/bin/
```
nathancoleman, line 104: Notably this now copies the fips version of envoy that **has** had `setcap` run on it

nathancoleman: @skpratt is there any specific testing I should do around this fips image?

pkg/envoy/proxy.go
```diff
@@ -392,7 +353,7 @@ func (p *Proxy) buildCommand(ctx context.Context, cfgPath string) *exec.Cmd {
 
 	args := append(
 		[]string{
-			"--config-path", cfgPath,
+			"--config-yaml", cfgYaml,
```
nathancoleman, line 356: Alongside the `setcap` changes applied in #238 , this is the meat of the change. Instead of writing the config to disk at `os.TempDir()` and then passing the filepath to the `envoy` process, the config yaml -- or JSON in our case -- is passed directly. JSON is allowed here according to the [envoy docs](https://www.envoyproxy.io/docs/envoy/latest/operations/cli#cmdoption-config-yaml).

pkg/envoy/proxy_test.go
```diff
@@ -15,11 +17,26 @@ import (
...
+				"initial_fetch_timeout": "0s",
+				"resource_api_version": "V3"
+			},
+			"lds_config": {
+				"ads": {},
+				"initial_fetch_timeout": "0s",
+				"resource_api_version": "V3"
+			},
+		}
+	}`)
```
nathancoleman, line 39: I aimed to test with some more robust json config vs. just "hello world" to ensure that the config retains its integrity when being passed through as a command line argument.

pkg/envoy/proxy_test.go
```diff
@@ -59,23 +83,27 @@ func TestProxy(t *testing.T) {
 		return true
 	}, 2*time.Second, 50*time.Millisecond)
 
+	// Check the output from fake-envoy to make sure we didn't hit an error case
+	require.Empty(t, envoyOut.String())
+	require.Empty(t, envoyErr.String())
```
nathancoleman, line 88: If the process hit an error case, such as a missing flag expected by `fake-envoy`, you would never know about in this test other than seeing the output file never being written.

With this change, we check for the output file with just an assert -- which allows the test to continue on -- and then make assertions about the output from Envoy so that errors become obvious when you're developing tests.

pkg/envoy/proxy_test.go
```diff
@@ -59,23 +83,27 @@ func TestProxy(t *testing.T) {
...
-	require.NoError(t, p.cmd.Process.Signal(syscall.Signal(0)))
+	assert.NoError(t, p.cmd.Process.Signal(syscall.Signal(0)))
 
 	// Ensure Kill kills and reaps the process.
 	require.NoError(t, p.Kill())
-
 	require.Eventually(t, func() bool {
-		return p.cmd.Process.Signal(syscall.Signal(0)) == os.ErrProcessDone
+		err := p.cmd.Process.Signal(syscall.Signal(0))
+		return errors.Is(err, os.ErrProcessDone)
```
nathancoleman, line 106: This was something the linter suggested as an improvement in order to support wrapped errors

pkg/envoy/testdata/fake-envoy
```diff
@@ -1,20 +1,20 @@
...
-# bootstrap config from the named pipe specified via `--config-path`, and writes
-# them to the file at `--test-output` (which is read and checked in the test).
+# bootstrap config specified via `--config-yaml`, and writes them to the file at
+# `--test-output` (which is read and checked in the test).
 # It then sleeps for 10 minutes to check we're correctly killing the process.
 
 set -e
 
-config_path=""
+config_yaml=""
```
nathancoleman, line 10: This fake envoy process will just stash the value of the `--config-yaml` argument it receives and write it to a file later so that the calling test can make assertions on it. The changes in this test are just around handling `--config-yaml` instead of `--config-path`.

nathancoleman (commented):
Personal review

Dockerfile
skpratt: Nope, it should be treated the same way as the non-FIPS one.

nathancoleman:
Applying the `do not merge` label while I do some deeper testing. Would love to get reviews on this in the meantime though and have no indication at time of writing that this solution won't work in all cases.

curtbushko (approved):
Let's try it out!

PR #254, opened 2023-09-01
Backport of NET-5186 Fix issue where consul-dataplane attempts to write to a read-only file system location into release/1.0.x

hc-github-team-consul-core:
## Backport

This PR is auto-generated from #253 to be assessed for backporting due to the inclusion of the label backport/1.0.

The below text is copied from the body of the original PR.

---

Running `setcap` on the consul-dataplane binary results in a process that cannot see the `TMPDIR` envar intended to influence where the envoy bootstrap config is written to. This is due to a previously-uknown-to-us side effect of glibc, described [here](https://serverfault.com/a/479081).

As a result, it attempts to write to the `/tmp` directory instead of `/consul/connect-inject` -- configured for injected sidecars [here](https://github.com/hashicorp/consul-k8s/blob/c064868527f9287df6a74d7a5a3eb4f3d3c3e40e/control-plane/connect-inject/webhook/consul_dataplane_sidecar.go#L81-L88) -- and gets a `read-only file system` error.

We previously considered dropping `setcap` for the `consul-dataplane` binary; however, the `envoy` binary that is forked from dataplane cannot itself have a capability that the consul-dataplane process does not have.

Instead, I've opted to keep `setcap net_bind_service` for both envoy and consul-dataplane. We can work around the fact that `TMPDIR` is invisible to `consul-dataplane` by not writing to the file system at all and using `envoy --config-yaml <yaml or json string>` where we were previously using `envoy --config-path <string>`.

---

<details>
<summary> Overview of commits </summary>

  - 518e8be445ea171806d4ef670d16a8f23f3567bd  - 1de6153702f0f89d3b0dab0dd2cd76254658b352  - fdf19892dbc7b2b1c5c99db6ecf592fdacdfd05f  - 039196332a7ad8745315c95c0ca17968f715cd15  - 268a2bc464b8abc2bc545e85c56fc1b313e44ea8  - 0e00abedfc0e5ec0a9b82af7b8a9abb2767009ae  - 45453c1a545fa6c14ea1a133d2d11e39935ae52f  - 54b21a4669c3a30c81e4d4ae33828e5f80e79cab  - 8a56a78f00bcdeb687ee95c5e8044d228df12693  - e96130d92ba19468b60fef8f715aa1cf0c954d90  - 87a5291e340a97b4ea2f80c83c496d17aaf8f67b 

</details>


nathancoleman (approved):
Matches source PR other than changes around FIPS envoy

PR #263, opened 2023-09-11
switch to hashicorp envoy image

skpratt:
Build run: https://github.com/hashicorp/consul-dataplane/actions/runs/6201904044

PR #268, opened 2023-09-22
Revert "switch to hashicorp envoy image (#263)"

jmurret:
This reverts commit 5eefda7e7ea5f44db41e72bb5a4c1474cec263a5 since hashicorp/envoy is not multi arch and is breaking consul-dataplane in use in acceptance tests.

### DIFF
--- a/.changelog/263.txt
+++ b/.changelog/263.txt
@@ -1,3 +1,0 @@
-```release-note:update
-Have consul dataplane use HashiCorp-published Envoy
-```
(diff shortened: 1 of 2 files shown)


nathancoleman (approved):

skpratt:
Hi John, I have a fix in the works for this with an ETA of later today - would it be possible to wait on this?

jmurret:
@skpratt sounds good!  thank you

PR #276, opened 2023-10-06
Update backport assistant to use merge commits

curtbushko:
Setting this environment variable causes the backport assistant to use the merge commit instead of individual commits when backporting a PR. This should prevent the assistant from trying to backport 100s of commits. Nomad uses this flag and it has helped a lot.

From the docs:

> BACKPORT_MERGE_COMMIT: When nonempty, backport-assistant will try to backport the merge commit instead of the individual commits that make of the PR. This will only work if you exclusively use the squash-merge commit strategy for PRs that get backported.

### DIFF
--- a/.github/workflows/backport-assistant.yml
+++ b/.github/workflows/backport-assistant.yml
@@ -16,7 +16,7 @@ jobs:
   backport:
     if: github.event.pull_request.merged
     runs-on: ubuntu-latest
-    container: hashicorpdev/backport-assistant:0.3.0
+    container: hashicorpdev/backport-assistant:0.3.5
     steps:
       - name: Run Backport Assistant for release branches
         run: |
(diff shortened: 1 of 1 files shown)


.github/workflows/backport-assistant.yml
```diff
@@ -16,12 +16,16 @@ jobs:
...
       - name: Run Backport Assistant for release branches
         run: |
           backport-assistant backport -merge-method=squash -gh-automerge
         env:
           BACKPORT_LABEL_REGEXP: "backport/(?P<target>\\d+\\.\\d+)"
           BACKPORT_TARGET_TEMPLATE: "release/{{.target}}.x"
+          # This forces the backport assistant to backport the merged commit
+          # instead of each commit individually. The environment variable
+          # just needs to exist for this to happen.
+          BACKPORT_MERGE_COMMIT: false
```
hashi-derek, line 30: Should we put in a value of `true` if the value simply needs to exist? Maybe I'm misunderstanding, but if the existence of `BACKPORT_MERGE_COMMIT` means it's true, then explicitly setting the value of false is super misleading.

hashi-derek: It's still super confusing in the description lol

hashi-derek (approved):

PR #281, opened 2023-10-06
Get correct docker tags for everything

curtbushko:
Bringing over #271


wilkermichael (approved):
Thanks for bringing this over!

PR #306, opened 2023-10-18
build(docker): always publish full and minor version tags for dev images

DanStough:
Always publishing full and minor-only tag versions. This accomplishes two goals:
1. For long running release branches, `.x`, always publish a minor-only tag so that tests use the latest version (e.g. `1.4-dev`)
2. For point release branches, `.0`, always publish the full version so we can fix this for testing purposes (e.g. `1.3.0`)

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -187,7 +187,10 @@ jobs:
       # This naming convention will be used ONLY for per-commit dev images
       - name: Set docker dev tag
         run: |
-          echo "dev_tag=${{ env.version }}" >> $GITHUB_ENV
+          echo "full_dev_tag=${{ env.version }}"
+          echo "full_dev_tag=${{ env.version }}" >> $GITHUB_ENV
+          echo "minor_dev_tag=$(echo ${{ env.version }}| sed -E 's/([0-9]+\.[0-9]+)\.[0-9]+(-[0-9a-zA-Z\+\.]+)?$/\1\2/')" 
+          echo "minor_dev_tag=$(echo ${{ env.version }}| sed -E 's/([0-9]+\.[0-9]+)\.[0-9]+(-[0-9a-zA-Z\+\.]+)?$/\1\2/')" >> $GITHUB_ENV
 
       - name: Docker Build (Action)
         if: ${{ !matrix.fips }}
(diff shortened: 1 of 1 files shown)


DanStough:
I expect the backports to fail but at least they will be a starting point.

.github/workflows/build.yml
```diff
@@ -187,7 +187,10 @@ jobs:
       # This naming convention will be used ONLY for per-commit dev images
       - name: Set docker dev tag
         run: |
-          echo "dev_tag=${{ env.version }}" >> $GITHUB_ENV
+          echo "full_dev_tag=${{ env.version }}"
+          echo "full_dev_tag=${{ env.version }}" >> $GITHUB_ENV
+          echo "minor_dev_tag=$(echo ${{ env.version }}| sed -E 's/([0-9]+\.[0-9]+)\.[0-9]+(-[0-9a-zA-Z\+]+)?$/\1\2/')" 
+          echo "minor_dev_tag=$(echo ${{ env.version }}| sed -E 's/([0-9]+\.[0-9]+)\.[0-9]+(-[0-9a-zA-Z\+]+)?$/\1\2/')" >> $GITHUB_ENV
```
zalimeni, line 193: Re: async conversation about making this consistent everywhere, I think we need to allow literal `.` as well [for FIPS](https://hub.docker.com/layers/hashicorppreview/consul-enterprise/1.16.0-dev-ent.fips1402-dev-ubi/images/sha256-471fd3d01063226f9b6e400aae8b3ad1262c95ae673b572ecb87e93bf86b52b8?context=explore) in `consul-enterprise`; might be simplest just to allow for that here and in other regexes too.

.github/workflows/build.yml
```diff
@@ -329,12 +344,12 @@ jobs:
     strategy:
       matrix:
         server:
-          - version: v1.15.0-dev
-            image: hashicorppreview/consul:1.15-dev
```
zalimeni, line 333: 1.15 will be LTS, and gets a patch next week - should probably keep for now?

zalimeni (approved):
LGTM, thank you!!

.github/workflows/build.yml
```diff
@@ -187,7 +187,10 @@ jobs:
       # This naming convention will be used ONLY for per-commit dev images
       - name: Set docker dev tag
         run: |
-          echo "dev_tag=${{ env.version }}" >> $GITHUB_ENV
+          echo "full_dev_tag=${{ env.version }}"
```
curtbushko, line 190: **praise:** I like that you echo out the tags... it will help with debugging in the future.

curtbushko (approved):
LGTM

.github/workflows/build.yml
DanStough: That's fair, I will update 👍

.github/workflows/build.yml
zalimeni: I think this line is where the `.` sneaks in (I remembered bc this bit me in some other situation during 1.16 RC): https://github.com/hashicorp/consul-enterprise/blob/d45b3bbf1085ed52de6015a6c780a9776bb500ab/.github/workflows/build.yml#L533

DanStough: Some of the other repos have the `.`, so I will add.

DanStough:
Backport assistant died. <cracks knuckles>.  Going in manually.

PR #341, opened 2023-11-07
chore: skaffold build experiment

DanStough:
Changes proposed in this PR:
- Adds a new make target and Dockerfile meant for dev builds or skaffold

How I've tested this PR:
[See this repo for usage](https://github.com/DanStough/blorp)

How I expect reviewers to test this PR: 👀

### DIFF
--- a/Dockerfile.dev
+++ b/Dockerfile.dev
@@ -1,3 +1,9 @@
+# DANGER: this dockerfile is experimental and could be modified/removed at any time.
+# A simple image for testing changes to consul-dataplane
+#
+# Meant to be used with the following make target
+# DEV_IMAGE=<name:tag> make skaffold
+
 FROM hashicorp/consul-dataplane as cache
 ARG TARGETARCH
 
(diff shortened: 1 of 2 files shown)


Dockerfile.dev
```diff
@@ -0,0 +1,4 @@
+FROM hashicorp/consul-dataplane as cache
```
zalimeni, line 1: Should we pin this to the latest minor line of tags so we're not relying on `latest` (which might get stale or accidentally overwritten w/ some random version)?

DanStough: IMHO, the `latest` usage is intentional. Since the dockerfile dependencies never really change between releases, it seems safe to use the newest possible version with binaries replaced. My thinking is that in the work case you build 1.0.0 with a newer baser image and dependency like `go-discover`, which should be a safe operation. Also, it's one less thing to update at release time :) 

Open to other cases I haven't thought about though.

zalimeni: Ah, TIL [CRT checks](https://go.hashi.co/ui/edit/crt-latest-docker) for the latest minor before updating the `latest` tag 👍🏻 SGTM!

zalimeni (approved):

PR #357, opened 2023-12-07
Increase max message receive size.

hashi-derek:
This commit changes the grpc-proxy internal to dataplane so that it can send / receive messages up to 50MB. Prior to this change, it would only allow 4MB payloads, which was insufficient for some proxies that has a very large number of upstreams. Whenever this limit was exceeded, the gRPC connection to the server would be terminated, causing a reconnect to occur, which would then repeat the issue all over again.

Because the underlying grpc-proxy library did not support configuring this field, I have had to refer to a custom fork of my own. Eventually if the corresponding change makes its way upstream, the go.mod file can point back to the original codebase.

It's worth noting that the lineage of this library is a bit confusing.

https://github.com/mwitkow/grpc-proxy is the original codebase, which contained a bug at the time it was added to dataplane. A different author had fixed the bug in their own fork, and I suppose it took a while to get it merged, so we pointed to their repo instead: github.com/adamthesax/grpc-proxy .

So rather than going two-forks-deep for this fix, I checked that the adamthesax bugfix had been merged upstream into mwitkow and forked from there. Based on code comparisons, they appear to be mostly the same, so this should be safe.


pkg/consuldp/xds.go
```diff
@@ -73,7 +73,9 @@ func (cdp *ConsulDataplane) setupXDSServer() error {
 	// However, we needed this fix (https://github.com/mwitkow/grpc-proxy/pull/62) which was available on the fork we are using.
 	// TODO: Switch to the main library once the fix is merged to keep upto date.
 	newGRPCServer := grpc.NewServer(
-		grpc.UnknownServiceHandler(proxy.TransparentHandler(cdp.director)),
+		// Increase the maximum message size due to large proxies sometimes exceeding the default 4MB limit.
+		grpc.MaxRecvMsgSize(50*1024*1024),
```
rboyer, line 77: Maybe move this 50MB into a constant since it's a magic number and you use it twice?

rboyer (approved):

PR #358, opened 2023-12-07
Backport of Increase max message receive size. into release/1.1.x

hc-github-team-consul-core:
## Backport

This PR is auto-generated from #357 to be assessed for backporting due to the inclusion of the label backport/1.1.

:rotating_light:
>**Warning** automatic cherry-pick of commits failed. If the first commit failed,
you will see a blank no-op commit below. If at least one commit succeeded, you
will see the cherry-picked commits up to, _not including_, the commit where
the merge conflict occurred.

The person who merged in the original PR is:
@hashi-derek
This person should manually cherry-pick the original PR into a new backport PR,
and close this one when the manual backport PR is merged in.

> merge conflict error: POST https://api.github.com/repos/hashicorp/consul-dataplane/merges: 409 Merge conflict []

The below text is copied from the body of the original PR.

---

This commit changes the grpc-proxy internal to dataplane so that it can send / receive messages up to 50MB. Prior to this change, it would only allow 4MB payloads, which was insufficient for some proxies that has a very large number of upstreams. Whenever this limit was exceeded, the gRPC connection to the server would be terminated, causing a reconnect to occur, which would then repeat the issue all over again.

Because the underlying grpc-proxy library did not support configuring this field, I have had to refer to a custom fork of my own. Eventually if the corresponding change makes its way upstream, the go.mod file can point back to the original codebase.

It's worth noting that the lineage of this library is a bit confusing.

https://github.com/mwitkow/grpc-proxy is the original codebase, which contained a bug at the time it was added to dataplane. A different author had fixed the bug in their own fork, and I suppose it took a while to get it merged, so we pointed to their repo instead: github.com/adamthesax/grpc-proxy .

So rather than going two-forks-deep for this fix, I checked that the adamthesax bugfix had been merged upstream into mwitkow and forked from there. Based on code comparisons, they appear to be mostly the same, so this should be safe.

Below is the comparison of my own custom changes against the upstream.
https://github.com/mwitkow/grpc-proxy/compare/master...hashi-derek:grpc-proxy:derekm/rename-packages

---

<details>
<summary> Overview of commits </summary>

  - f6d6fd1c9ed51384f3e93eb7f4ccf8fa4628f295 

</details>


hashi-derek:
Closing due to failed cherry-pick on backport which locks up the CLA.

PR #372, opened 2023-12-20
[NET-5263] Pass query to Envoy when proxying metrics requests

zalimeni:
### Changes proposed in this PR

* Allow passthrough of merged metrics endpoint query params to Envoy's stats endpoint to enable filtering of metrics
* Prevent query in user-defined scrape path from breaking path matching for the merged metrics endpoint

### Considerations

An alternative approach could use explicit config in `consul-dataplane` and the `consul-k8s` Helm chart to set static query parameters to append to Envoy stats requests. That would require more work to implement, and would also limit the ability of multiple consumers (e.g. old + new Prometheus config, or a human/other scrape process) from using different queries to suit their needs.

We could also implement an explicit allow-list for query params. After reviewing the available options in the [docs](https://www.envoyproxy.io/docs/envoy/latest/operations/admin#get--stats?format=prometheus&usedonly), it seems reasonable to allow users to interact freely with that specific endpoint to take advantage of new features Envoy may offer in the future. While an allow-list could also be added via configuration, this doesn't seem necessary at the moment, and we already don't limit the `envoy_stats_bind_addr` (`/stats`) endpoint today. That said, open to discussion if folks disagree.

### Testing

* Tested locally in `kind` cluster:
```shell
❯ k exec $(get-pod-name api) -c api -- curl -s 'localhost:20200/custom/stats/prometheus' | wc -l
    2364
❯ k exec $(get-pod-name api) -c api -- curl -s 'localhost:20200/custom/stats/prometheus?usedonly=true' | wc -l
     888
```

### Links

* Envoy admin `/stats` endpoint [docs](https://www.envoyproxy.io/docs/envoy/latest/operations/admin#get--stats?format=prometheus&usedonly) (note that `/stats/prometheus` is an alias for `?format=prometheus`, and other `/stats` query params work for both endpoints)

### DIFF
--- a/.changelog/372.txt
+++ b/.changelog/372.txt
@@ -1,0 +1,6 @@
+```release-note:improvement
+Propagate merged metrics request query params to Envoy to enable metrics filtering.
+```
+```release-note:bug
+Exclude Prometheus scrape path query params from Envoy path match s.t. it does not break merged metrics request routing.
+```
--- a/internal/bootstrap/bootstrap_config.go
+++ b/internal/bootstrap/bootstrap_config.go
@@ -16,6 +16,8 @@ import (
 	"path"
 	"strings"
 	"text/template"
+
+	"github.com/hashicorp/go-hclog"
 )
 
 const (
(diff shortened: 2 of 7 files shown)


DanStough (approved):
LGTM

david-yu:
@zalimeni No need to backport here, I think we can just introduce this as part of dataplane 1.4.0.

zalimeni:
> @zalimeni No need to backport here, I think we can just introduce this as part of dataplane 1.4.0.

@david-yu SGTM - thanks!

zalimeni:
Added changelog

PR #384, opened 2024-01-12
Update go to 1.21.6 and use .go-version file everywhere

curtbushko:
### Changes proposed in this PR ###  
- Upgrade Go to 1.21.6
  - This is needed in advance of the [EOL of Go 1.20 ~Feb. 1](https://endoflife.date/go). The ideal timing is before our next patch release scheduled for the end of January.
- Updated makefile, github actions and Dockerfile to use `.go-version` file as a single source of truth.
- Note: `hashicorp/actions-docker-build` needed to be updated to handle docker build arguments

### How I've tested this PR ###

- builds and tests

### How I expect reviewers to test this PR ###

:eyes:

### Checklist ###
- [ ] Tests added
- [x] [CHANGELOG entry added](https://github.com/hashicorp/consul-k8s/blob/main/CONTRIBUTING.md#adding-a-changelog-entry)


Dockerfile
```diff
@@ -9,6 +9,7 @@
 
 # envoy-binary pulls in the latest Envoy binary, as Envoy don't publish
 # prebuilt binaries in any other form.
+ARG GOLANG_VERSION
```
zalimeni, line 12: ~ Is this needed here?

Makefile
```diff
@@ -1,20 +1,21 @@
...
 BIN_NAME     ?= $(PRODUCT_NAME)
 GOPATH       ?= $(shell go env GOPATH)
 GOBIN        ?= $(GOPATH)/bin
 
 # Get local ARCH; on Intel Mac, 'uname -m' returns x86_64 which we turn into amd64.
 # Not using 'go env GOOS/GOARCH' here so 'make docker' will work without local Go install.
 ARCH     ?= $(shell A=$$(uname -m); [ $$A = x86_64 ] && A=amd64; echo $$A)
-OS       ?= $(shell uname | tr [[:upper:]] [[:lower:]])
+# Only build for linux so that building the docker image works on M1 Macs.
+OS       ?= linux
```
zalimeni, line 12: Is this effectively equivalent or is there anywhere else this got used other than the removed line below?

zalimeni (approved):
LGTM! Two small questions but approving to unblock once those close

Makefile
curtbushko: Some things were failing to build for me on M1 because the OS does not match what we should build. We should always be building 'linux' as the OS. 

An example of [us forcing it in consul-k8s](https://github.com/hashicorp/consul-k8s/blob/main/Makefile#L44)

Dockerfile
curtbushko: It needed to be added back.

zalimeni: Ah, I think multi-stage needs it declared up top https://stackoverflow.com/questions/53681522/share-variable-in-multi-stage-dockerfile-arg-before-from-not-substituted

zalimeni (approved):

PR #370, opened 2023-12-19
Add exporting of Envoy metrics to HCP for linked clusters

jjti:
## Overview

This builds off the work of the Consul PR: https://github.com/hashicorp/consul/pull/20257

We want to let users send Envoy metrics to HCP for use in service-level dashboards and coming service topology map features. We also want to support that without requiring that users deploy a dedicated [Consul Telemetry Collector](https://developer.hashicorp.com/hcp/docs/consul/monitor/consul-central/observability/telemetry-collector). 

The behavior added to Consul Dataplane in this PR is:
1. Watch for changes to TelemetryState in Consul via the resource API
2. On creation of a TelemetryState resource, create an HTTP client to push metrics to HCP's Consul metrics endpoint
3. Start periodically (minutely) scraping Envoy's stats endpoint
4. Convert those Prometheus metrics to OTLP metrics
5. POST the metrics to HCP

Screenshot 2024-01-26 at 11 21 03 AM

More details are available in the RFC: https://docs.google.com/document/d/1gh3_zUtdIDeDPIH7z3d0-kkx1scB2oZLi7yuYg88PK8

### Unit/integration Testing

```
$go test -timeout 30s -tags integration github.com/hashicorp/consul-dataplane/pkg/hcp/telemetry

ok  	github.com/hashicorp/consul-dataplane/pkg/hcp/telemetry	0.788s	coverage: 90.7% of statements
```

```
go test -timeout 30s -tags integration github.com/hashicorp/consul-dataplane/pkg/hcp/telemetry/otlphttp

ok  	github.com/hashicorp/consul-dataplane/pkg/hcp/telemetry/otlphttp	0.651s	coverage: 88.9% of statements
```

### E2E Testing with HCP

To test this with real values I:
1. created a resource in consul-core by calling the resource API with grpcurl (using a real endpoint, filters, labels, and service principal)
2. validated that metrics were pushed through the HCP metrics endpoint and available for querying

```bash
# create a resource manually
grpcurl -d @ \
  -plaintext \
  -protoset pkg/consul.protoset \
  127.0.0.1:8502 \
  hashicorp.consul.resource.ResourceService.Write \
<<EOF
  {
    "resource": {
      "id": {
        "type": {
          "group": "hcp",
          "group_version": "v2",
          "kind": "TelemetryState"
        },
        "name": "global"
      },
      "metadata": {
        "consul.io/hcp/telemetry-state/debug/skip-deletion": "true"
      },
      "data": {
[...]

PR #406, opened 2024-02-01
Bump consul/proto-public and create mocks

jjti:
### Changes

This pulls in the latest public proto specifications from consul's proto-public package. We need this for a subsequent PR that references the newly created `TelemetryState` struct: https://github.com/hashicorp/consul-dataplane/pull/370

Bumping `proto-public` breaks the tests in `pkg/dns/dns_test.go` because the referenced mock is no longer exported. To get around that, while keeping the dns changes largely unchanged, a new mockery mock is created using the newly added make target below:

```make
# This generates mocks against public proto packages in consul. At the time of writing,
# only the dns and resource packages are used in consul-dataplane so only mocks for their
# interfaces are generated here.
.PHONY: mocks
mocks:
	for pkg in pbdns pbresource; do \
		mockery --srcpkg=github.com/hashicorp/consul/proto-public/$$pkg --output ./internal/mocks/$${pkg}mock --outpkg $${pkg}mock --case underscore --all; \
	done
```

The `pbresource` mocks are also added here as they are used in the next PR.

### Testing

These changes were pulled out of the follow-up PR where they were tested by building + deploying a custom build of consul-dataplane using these changes.


pkg/dns/dns_test.go
```diff
@@ -40,7 +41,7 @@ func genRandomBytes(size int) (blk []byte) {
 }
 
 func (s *DNSTestSuite) Test_DisabledServer() {
-	mockedDNSConsulClient := pbdns.NewMockDNSServiceClient(s.T())
```
jjti, line 43: `NewMockDNSServiceClient` no longer exists in `github.com/hashicorp/consul/proto-public/pbdns`

wilkermichael:
@jjti qq: is this meant to be supported in the next release (Consul 1.18/dataplane 1.14)?

jjti:
> @jjti qq: is this meant to be supported in the next release (Consul 1.18/dataplane 1.14)?

Yes and that's correct

jjti:
> @jjti qq: is this meant to be supported in the next release (Consul 1.18/dataplane 1.14)?

@wilkermichael yes that's correct

wilkermichael:
Ok, we'll need to add a backport label, just waiting for it to be created. The release process is still ongoing for dataplane

.changelog/406.txt
DanStough, line 1: IMHO we don't need a changelog since this is an internal, intermediate change and you'll have a feature one for #370

DanStough (approved):
LGTM and thanks for this fix! See my comment about the changelog.

One rhetorical point that has nothing to do with this PR:  it's weird we got rid of the mocks to cut down on dependencies in `proto-public`, but we ended up pulling in a ton of K8s dependencies. 🤷

.changelog/406.txt
jjti: Makes sense! Removed the changelog

jjti:
> > One rhetorical point that has nothing to do with this PR: it's weird we got rid of the mocks to cut down on dependencies in `proto-public`, but we ended up pulling in a ton of K8s dependencies. 🤷
> 
> I hear you @DanStough and that's something I'll call out here: it appears to increase the consul-dataplane binary size ~50% (19M > 29M):
> 
> ```shell
> # josh @ josh-XWGWHJ0JV2 in ~/GitHub/hashicorp/consul-dataplane on git:bump-protos
> $ make bin
> GOARCH=arm64 GOOS=linux CGO_ENABLED=0 go build -trimpath -buildvcs=false -ldflags="-X github.com/hashicorp/consul-dataplane/pkg/version.GitCommit=6ddd5cc" -o dist/linux/arm64/consul-dataplane ./cmd/consul-dataplane
> 
> # josh @ josh-XWGWHJ0JV2 in ~/GitHub/hashicorp/consul-dataplane on git:jjti/bump-protos o [16:26:52] 
> $ ll dist/linux/arm64/consul-dataplane                                    
> -rwxr-xr-x  1 josh  staff    29M Feb  2 16:22 dist/linux/arm64/consul-dataplane
> 
> # josh @ josh-XWGWHJ0JV2 in ~/GitHub/hashicorp/consul-dataplane on git:main o [16:33:17] 
> $ make bin
> GOARCH=arm64 GOOS=linux CGO_ENABLED=0 go build -trimpath -buildvcs=false -ldflags="-X github.com/hashicorp/consul-dataplane/pkg/version.GitCommit=6ddd5cc" -o dist/linux/arm64/consul-dataplane ./cmd/consul-dataplane
> 
> # josh @ josh-XWGWHJ0JV2 in ~/GitHub/hashicorp/consul-dataplane on git:main o [16:33:26] 
> $ ll dist/linux/arm64/consul-dataplane
> -rwxr-xr-x  1 josh  staff    19M Feb  2 16:33 dist/linux/arm64/consul-dataplane
> ```
> 
> Similar/same increase image for the image: 139MB to 150MB

There was a discussion about this in Slack. The gateway team will look at removing the generated code import of k8s to avoid this increase. Ie this increase in size, from this PR, is temporary until the next tagged release of `consul/proto-public` after which we'll be back to the size we'd been at before

PR #416, opened 2024-02-10
[NET-7948] security: update Envoy to 1.28.1

david-yu:
Bump to 1.28.1 on `main` and `release/1.4.x`.

### DIFF
--- a/.changelog/416.txt
+++ b/.changelog/416.txt
@@ -1,0 +1,6 @@
+```release-note:improvement
+Update Envoy version from 1.27 to 1.28
+```
+```release-note:security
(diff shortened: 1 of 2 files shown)


zalimeni:
@david-yu it looks like we might have missed the Envoy 1.28.1 bump previously here in `consul-dataplane`, I'm guessing bc that upgrade was delayed due to some fixes in `consul`. `main` and `release/1.4.x` should be on that version now, so updating this PR accordingly; we can open a separate one for 1.3.x to move to Envoy 1.27.3. Correct me if that sounds off to you.

zalimeni:
Note: before the next release, we'll need to fast-follow this w/ a `hashicorp/envoy-fips` bump once that's available

zalimeni (approved):
Updated to target 1.28.1 - @david-yu I'm approving but LMK if this looks off in any way based on above comments before merging

david-yu:
Thanks I may have forgot a previous conversation where we agreed to move to 1.28.x. I also was not sure if this was not bumped for a reason. Either way, it makes sense to do so that we do not risk running into an EOL for Consul 1.17.x later down the line.

david-yu:
LGTM

PR #417, opened 2024-02-10
[NET-7948] security: update Envoy to 1.26.7

david-yu:
Bump from Envoy 1.26.6 to Envoy 1.26.7


david-yu:
Thank you!

zalimeni:
Note: before the next release, we'll need to fast-follow this w/ a `hashicorp/envoy-fips` bump once that's available

PR #421, opened 2024-02-12
[NET-7948] security: update Envoy to 1.27.3 (release/1.3.x)

zalimeni:
Bump from Envoy 1.27.2 to Envoy 1.27.3


zalimeni:
Note: before the next release, we'll need to fast-follow this w/ a hashicorp/envoy-fips bump once that's available

david-yu (approved):

PR #438, opened 2024-02-15
Stage release v1.4.0

DanStough:
Stage the 1.4.0 branch for code freeze. Proto-public updates to come later after finalized by consul.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,24 @@
+## 1.4.0 (February 28, 2024)
+
+SECURITY:
+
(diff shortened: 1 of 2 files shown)


CHANGELOG.md
```diff
@@ -1,3 +1,24 @@
+## 1.4.0 (February 28, 2024)
```
zalimeni, line 1: Assume we'll just regenerate this later by re-running the make target (if backports happen)?

zalimeni (approved):

CHANGELOG.md
DanStough: Yeah, that's a good call. TBH I totally forgot about it reading the instructions. I'm inclined to say keeping it is harmless if we don't have changes but open to feedback.

PR #522, opened 2024-05-24
build: prepare for 1.6.0 dev

DanStough:


zalimeni (approved):

DanStough:
I updated this PR to include 1.20 testing since it should be compatible with`main` dataplane.

.github/workflows/build.yml
```diff
@@ -355,12 +355,14 @@ jobs:
         server:
           - version: v1.15.0-dev
             image: hashicorppreview/consul:1.15-dev
-          - version: v1.16.0-dev
-            image: hashicorppreview/consul:1.16-dev
           - version: v1.17.0-dev
             image: hashicorppreview/consul:1.17-dev
```
zalimeni, line 359: I think we can drop this one as well, since 1.20 will have an active set of 1.20,1.19,1.18 LTS,1.15 LTS

PR #524, opened 2024-05-28
Add Consul 1.19 int test to 1.5.x

zalimeni:
In addition to `main`, we should be testing these versions together on the release branch.

Drop 1.16, which is retired as of 1.5.

See https://github.com/hashicorp/consul-dataplane/pull/522 for `main` updates.


dhiaayachi (approved):
LGTM!

zalimeni:
Note: I expect this to fail due to int failures in the linked `main` PR; regardless, we need to fix both.

PR #534, opened 2024-06-10
docs: fix missing changelog entries up to 1.4.2

DanStough:
We were missing some entries, and some were mislabeled. I also modified the dates to match the GitHub releases.

I am going to do manual backports since the context is different for each branch and they diverge in different ways.


zalimeni (approved):

zalimeni:
Thank you!

PR #545, opened 2024-06-12
[NET-10001] test: replace deprecated google/pause container

zalimeni:
This container is 9+ years unmaintained and no longer pullable by default due to using a deprecated manifest format. Replacing with a common alternative pause image.

I'll backport this manually to 1.5.0 as well.


dhiaayachi (approved):
🚢

DanStough (approved):

PR #560, opened 2024-06-13
ci: add conditional skip for docs changes

zalimeni:
[Similar to `consul`](https://github.com/hashicorp/consul/blob/e9c983f36107b044fcd889f738b3d5ba2b922e7a/.github/workflows/go-tests.yml#L33), skip build and tests if we're only changing docs.

This should help speed up releases and daily work quick-fixes.

```shell
❯ SKIP_CHECK_BRANCH=zalimeni/conditional-skip-ci GITHUB_OUTPUT="../GITHUB_OUTPUT" .github/scripts/check_skip_ci.sh
checking file: .changelog/100.txt
checking file: .github/scripts/check_skip_ci.sh
checking file: README.md
Changes detected in only documentation files - skipping tests and build
```
After adding a code change:
```shell
❯ SKIP_CHECK_BRANCH=zalimeni/conditional-skip-ci GITHUB_OUTPUT="../GITHUB_OUTPUT" .github/scripts/check_skip_ci.sh
checking file: .changelog/100.txt
checking file: .github/scripts/check_skip_ci.sh
checking file: README.md
checking file: cmd/consul-dataplane/main.go
non-skippable file changed: cmd/consul-dataplane/main.go
Changes detected in non-documentation files - will not skip tests and build
```

### DIFF
--- a/.github/scripts/check_skip_ci.sh
+++ b/.github/scripts/check_skip_ci.sh
@@ -1,0 +1,49 @@
+#!/bin/bash
+# Copyright (c) HashiCorp, Inc.
+# SPDX-License-Identifier: MPL-2.0
+
+set -euo pipefail
+
+# Get the list of changed files
+# Using `git merge-base` ensures that we're always comparing against the correct branch point.
+#For example, given the commits:
+#
+# A---B---C---D---W---X---Y---Z # origin/main
+#             \---E---F         # feature/branch
+#
+# ... `git merge-base origin/$SKIP_CHECK_BRANCH HEAD` would return commit `D`
+# `...HEAD` specifies from the common ancestor to the latest commit on the current branch (HEAD)..
+files_to_check=$(git diff --name-only "$(git merge-base origin/$SKIP_CHECK_BRANCH HEAD~)"...HEAD)
+
+# Define the directories to check
+skipped_directories=("_doc/" ".changelog/")
+
+# Loop through the changed files and find directories/files outside the skipped ones
+files_to_check_array=($files_to_check)
+for file_to_check in "${files_to_check_array[@]}"; do
+	file_is_skipped=false
+	echo "checking file: $file_to_check"
+
+	# Allow changes to:
+	# - This script
+	# - Files in the skipped directories
+	# - Markdown files
+	for dir in "${skipped_directories[@]}"; do
+		if [[ "$file_to_check" == */check_skip_ci.sh ]] ||
+		   [[ "$file_to_check" == "$dir"* ]] ||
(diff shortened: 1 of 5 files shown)


.github/scripts/check_skip_ci.sh
```diff
@@ -0,0 +1,51 @@
...
+#
+# ... `git merge-base origin/$SKIP_CHECK_BRANCH HEAD` would return commit `D`
+# `...HEAD` specifies from the common ancestor to the latest commit on the current branch (HEAD)..
+files_to_check=$(git diff --name-only "$(git merge-base origin/$SKIP_CHECK_BRANCH HEAD~)"...HEAD)
+
+# Define the directories to check
+skipped_directories=("_doc/" ".changelog/")
+
+# Loop through the changed files and find directories/files outside the skipped ones
+files_to_check_array=($files_to_check)
```
zalimeni, line 22: We were not converting to an array in the original script - it happened to still work due to multiline matches, but this allows us to print individual debug logs more accurately

zalimeni:
I was thinking the skipped jobs would be considered passed, but I must be forgetting how this works. I'll get that sorted out so merge is allowed.

zalimeni:
See check here: https://github.com/hashicorp/consul-dataplane/actions/runs/9504971128/job/26198781967#step:3:9

zalimeni:
@nathancoleman when you get a sec, could I get a follow-up review of this change too? (basically the same as https://github.com/hashicorp/consul-dataplane/pull/560 apart from exempt directory list and integrated workflows)

.github/workflows/build.yml
```diff
@@ -401,9 +410,10 @@ jobs:
 
   integration-tests-success:
     needs:
+      - conditional-skip
       - integration-tests
     runs-on: ubuntu-latest
-    if: always()
+    if: always() && needs.conditional-skip.outputs.skip-ci != 'true'
```
DanStough, line 416: So the action won't get stuck because this job is required and there are cases where we won't run it, correct?

zalimeni: Exactly. This follows the pattern in `consul` [here](https://github.com/hashicorp/consul/blob/7a19d2e7a48ffc7a6b16889adad63b121e98d82c/.github/workflows/test-integrations.yml#L599)

zalimeni:
[Removed](https://github.com/hashicorp/consul-dataplane/compare/f663c269eee50063c3d305c4a71ef696a296338e..63d4bacb4c40546cb333c7b903b14a38111ccec9) dead env var accidentally left in from `consul` script

DanStough (approved):
Thank you!

PR #566, opened 2024-06-28
remove macos runners from buildworkflow

NicoletaPopoviciu:
remove macos runners from buildworkflow

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -100,6 +100,7 @@ jobs:
             ${{ matrix.env }} go build -tags=${{ matrix.gotags }} -trimpath -buildvcs=false -ldflags="-X github.com/hashicorp/consul-dataplane/pkg/version.GitCommit=${GITHUB_SHA::8}" -o $BIN_PATH ./cmd/$BIN_NAME
 
       - name: Copy license file
+        if: ${{ matrix.goos == 'linux' }}
         env:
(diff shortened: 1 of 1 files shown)


.github/workflows/build.yml
DanStough, line 1: Similar comment here as the PR in `consul-aws`. I think we can remove the `build-darwin` step and add entries in `build-linux` around line 73. In this case we will also need to modify the `Copy license file` step so that is only runs on Linux.

DanStough (changes requested):

.github/workflows/build.yml
```diff
@@ -73,6 +73,8 @@ jobs:
           - {goos: "linux", goarch: "amd64"}
           - {goos: "linux", goarch: "amd64", gotags: "fips", env: "CGO_ENABLED=1 GOEXPERIMENT=boringcrypto", fips: "+fips1402", pkg_suffix: "-fips" }
           - {goos: "linux", goarch: "arm64", gotags: "fips", env: "CGO_ENABLED=1 GOEXPERIMENT=boringcrypto CC=aarch64-linux-gnu-gcc", fips: "+fips1402", pkg_suffix: "-fips" }
+          - {goos: "darwin", goarch: "arm"}
```
DanStough, line 76: This was a mistake in my original consul PR:
```suggestion
          - {goos: "darwin", goarch: "amd64"}
```

DanStough (changes requested):

DanStough (approved):

PR #593, opened 2024-07-22
NET-10480 - fix setting namespace and partition setting in dns-proxy mode

jmurret:
This fixes how namespace and partition get set when in dns-proxy mode.  The prior attempt at this moved starting the DNS above starting envoy.  What was not clear was that starting envoy set `resolvedProxyConfig` in state so it was not set when DNS started and dns-proxy always started with the default namespace and partition.  In the process, i also got rid of using state to pass parameters and just used the function signature to have the appropriate parameters.

### DIFF
--- a/pkg/consuldp/bootstrap.go
+++ b/pkg/consuldp/bootstrap.go
@@ -26,7 +26,7 @@ const (
 	defaultAdminAccessLogsPath = os.DevNull
 )
 
-// bootstrapConfig generates the Envoy bootstrap config in JSON format.
+// getBootstrapParams makes a call using the service client to get the bootstrap params for eventually getting the Envoy bootstrap config.
 func (cdp *ConsulDataplane) getBootstrapParams(ctx context.Context) (*pbdataplane.GetEnvoyBootstrapParamsResponse, error) {
 	svc := cdp.cfg.Proxy
 
--- a/pkg/consuldp/bootstrap.go
+++ b/pkg/consuldp/bootstrap.go
@@ -27,9 +27,8 @@ const (
 )
 
 // bootstrapConfig generates the Envoy bootstrap config in JSON format.
-func (cdp *ConsulDataplane) bootstrapConfig(ctx context.Context) (*bootstrap.BootstrapConfig, []byte, error) {
+func (cdp *ConsulDataplane) getBootstrapParams(ctx context.Context) (*pbdataplane.GetEnvoyBootstrapParamsResponse, error) {
 	svc := cdp.cfg.Proxy
-	envoy := cdp.cfg.Envoy
 
 	req := &pbdataplane.GetEnvoyBootstrapParamsRequest{
 		ServiceId: svc.ProxyID,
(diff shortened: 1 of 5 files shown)


pkg/consuldp/bootstrap.go
```diff
@@ -27,9 +27,8 @@ const (
 )
 
 // bootstrapConfig generates the Envoy bootstrap config in JSON format.
-func (cdp *ConsulDataplane) bootstrapConfig(ctx context.Context) (*bootstrap.BootstrapConfig, []byte, error) {
+func (cdp *ConsulDataplane) getBootstrapParams(ctx context.Context) (*pbdataplane.GetEnvoyBootstrapParamsResponse, error) {
```
jmurret, line 30: dns proxy requires bootstrap params to set the namespace and partition.....but it does not require the bootstrap config itself.  So we are splitting this into two functions, so that `startDNSProxy()` can get require bootstrap params in its signature rather than just assuming it is already set on the struct.

pkg/consuldp/bootstrap.go
```diff
@@ -27,9 +27,8 @@ const (
 )
 
 // bootstrapConfig generates the Envoy bootstrap config in JSON format.
```
jmurret, line 29: TODO: change comment

pkg/consuldp/bootstrap.go
```diff
@@ -50,16 +49,17 @@ func (cdp *ConsulDataplane) bootstrapConfig(ctx context.Context) (*bootstrap.Boo
...
-		ProxyID:   cdp.cfg.Proxy.ProxyID,
-		Namespace: rsp.Namespace,
-		Partition: rsp.Partition,
-	}
+	return rsp, nil
+}
+
+// bootstrapConfig generates the Envoy bootstrap config in JSON format.
+func (cdp *ConsulDataplane) bootstrapConfig(
+	bootstrapParams *pbdataplane.GetEnvoyBootstrapParamsResponse) (*bootstrap.BootstrapConfig, []byte, error) {
```
jmurret, line 60: A prior call will already have gotten bootstrap params for this call as well as `startDNSProxy()`

pkg/consuldp/consul_dataplane.go
```diff
@@ -180,8 +180,15 @@ func (cdp *ConsulDataplane) Run(ctx context.Context) error {
 	cdp.dpServiceClient = pbdataplane.NewDataplaneServiceClient(state.GRPCConn)
 
 	doneCh := make(chan error)
+	bootstrapParams, err := cdp.getBootstrapParams(ctx)
```
jmurret, line 183: get bootstrap params separately so it can be passed into `startDNSProxy()` and `bootstrapConfig()` calls separately.  This gets rid of relying on dataplane's `resolvedProxyConfig` to have been set prior and also the code reader and linter know this function relies on bootstrap params having been retrieved.

zalimeni (approved):
Thanks for refactoring this and adding test + log cover 👍🏻

PR #605, opened 2024-08-30
add security scan for each built image

dhiaayachi:


dhiaayachi:
We already have image scan as part of the CRT build.

PR #606, opened 2024-09-03
Update release 1.5.3

dhiaayachi:


pkg/version/version.go
```diff
@@ -17,7 +17,7 @@ var (
 	//
 	// Version must conform to the format expected by github.com/hashicorp/go-version
 	// for tests to work.
-	Version = "1.5.2"
+	Version = "1.5.3"
```
zalimeni, line 20: I think we want this to be 1.5.4 now, since we've just released 1.5.3 (made more confusing w/ the double bump)
```suggestion
	Version = "1.5.4"
```

PR #651, opened 2024-10-17
Update version.go to reflect next release

nathancoleman:


pkg/version/version.go
```diff
@@ -17,7 +17,7 @@ var (
 	//
 	// Version must conform to the format expected by github.com/hashicorp/go-version
 	// for tests to work.
-	Version = "1.6.0"
+	Version = "1.7.0"
```
jm96441n, line 20: should this be `1.7.0-dev`?

nathancoleman: the `dev` prerelease tag is below

jm96441n (approved):

PR #652, opened 2024-10-17
[Experimental] Include both privileged and non-privileged binaries in Docker image

nathancoleman:
Currently, the `NET_BIND_SERVICE` capability is added to the dataplane (and Envoy) entrypoints since we cannot know at build time whether a non-root user at runtime will attempt to bind to a privileged port. The reason this is the default behavior was to avoid a breaking change in ingress gateways, which had previously supported binding to privileged ports; however, there are still many use cases where it's preferable to avoid this capability when it isn't needed.

This change introduces a second set of entrypoints to the container that _do not_ have the `NET_BIND_SERVICE` capability. In this way, we can support both use cases:
- Ingress gateways, which have historically supported binding to privileged ports
- All other use cases (sidecar and API, mesh, terminating gateways) which _do not_ need to support binding to privileged ports

The container will default to the non-privileged entrypoint, and ingress gateway deployments that do want to support binding to privileged ports will need to override the entrypoint to `privileged-consul-dataplane` and provide a path to `privileged-envoy`. This way, no deployment needs the `NET_BIND_SERVICE` capability unless they are overriding the entrypoint in order to support binding to privileged ports.

See https://github.com/hashicorp/consul-k8s/pull/4394 for an example of how this works in consul-k8s.

> [!WARNING]
> We still need to determine the impact – if any – of a change like this for VM and Nomad deployments of consul-dataplane

### DIFF
--- a/cmd/consul-dataplane/main.go
+++ b/cmd/consul-dataplane/main.go
@@ -100,6 +100,7 @@ func init() {
 	IntVar(flags, &flagOpts.dataplaneConfig.Envoy.Concurrency, "envoy-concurrency", "DP_ENVOY_CONCURRENCY", "The number of worker threads that Envoy uses.")
 	IntVar(flags, &flagOpts.dataplaneConfig.Envoy.DrainTimeSeconds, "envoy-drain-time-seconds", "DP_ENVOY_DRAIN_TIME", "The time in seconds for which Envoy will drain connections.")
 	StringVar(flags, &flagOpts.dataplaneConfig.Envoy.DrainStrategy, "envoy-drain-strategy", "DP_ENVOY_DRAIN_STRATEGY", "The behaviour of Envoy during the drain sequence. Determines whether all open connections should be encouraged to drain immediately or to increase the percentage gradually as the drain time elapses.")
+	StringVar(flags, &flagOpts.dataplaneConfig.Envoy.ExecutablePath, "envoy-executable-path", "DP_ENVOY_EXECUTABLE_PATH", "Path to the Envoy executable to run. Defaults to the ")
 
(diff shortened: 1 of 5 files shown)


jm96441n:
this shouldn't affect nomad/VM deployments as those don't use dataplane

cmd/consul-dataplane/main.go
```diff
@@ -100,6 +100,7 @@ func init() {
 	IntVar(flags, &flagOpts.dataplaneConfig.Envoy.Concurrency, "envoy-concurrency", "DP_ENVOY_CONCURRENCY", "The number of worker threads that Envoy uses.")
 	IntVar(flags, &flagOpts.dataplaneConfig.Envoy.DrainTimeSeconds, "envoy-drain-time-seconds", "DP_ENVOY_DRAIN_TIME", "The time in seconds for which Envoy will drain connections.")
 	StringVar(flags, &flagOpts.dataplaneConfig.Envoy.DrainStrategy, "envoy-drain-strategy", "DP_ENVOY_DRAIN_STRATEGY", "The behaviour of Envoy during the drain sequence. Determines whether all open connections should be encouraged to drain immediately or to increase the percentage gradually as the drain time elapses.")
+	StringVar(flags, &flagOpts.dataplaneConfig.Envoy.ExecutablePath, "envoy-executable-path", "DP_ENVOY_EXECUTABLE_PATH", "Path to the Envoy executable to run. Defaults to the ")
```
zalimeni, line 103: ~ Looks like this needs a bit more text

zalimeni:
This seems like a great idea! Excited to see it happen. Hoping/expecting the constraints around `consul-dataplane` are acceptable.

cmd/consul-dataplane/main.go
jm96441n: good catch!

srahul3 (approved):
✅ LGTM

PR #684, opened 2025-02-11
Net 1783 Add PostStart hook to connect inject

nitin-sachdev-29:
Net 1783 Add PostStart hook to connect inject.
cloning from branch net-1783-poststart_hook


cmd/consul-dataplane/graceful_startup.go
```diff
@@ -0,0 +1,33 @@
...
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"time"
+)
+
+//const gracefulPort = "20600"
+
+func RunGracefulStartup(gracefulStartupPath string, gracefulPort int) error {
```
sreeram77, line 12: godoc is missing for exported func.

cmd/consul-dataplane/graceful_startup.go
```diff
@@ -0,0 +1,33 @@
+package main
+
+import (
+	"context"
+	"fmt"
+	"net/http"
+	"time"
+)
+
+//const gracefulPort = "20600"
```
sreeram77, line 10: Commented code should be removed.

cmd/consul-dataplane/main.go
```diff
@@ -146,7 +146,18 @@ func validateFlags() {
 }
 
 func run() error {
-	err := flags.Parse(os.Args[1:])
+	// Shift arguments by one if subcommand is the first argument.
+	subcommand := os.Args[1]
+	var arguments []string
+	switch subcommand {
+	case "graceful-startup":
```
sreeram77, line 153: Use const instead of hard-coding string.

cmd/consul-dataplane/graceful_startup.go
```diff
@@ -0,0 +1,33 @@
...
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return nil
+		case <-ticker.C:
+			resp, err := http.Get(fmt.Sprintf("http://localhost:%d%s", gracefulPort, gracefulStartupPath))
+			if err != nil {
+				fmt.Println(err.Error())
```
sreeram77, line 25: Is there a reason to use fmt instead of logger for logging errors?

pkg/consuldp/lifecycle.go
```diff
@@ -107,6 +107,10 @@ func (m *lifecycleConfig) startLifecycleManager(ctx context.Context) error {
 
 	// Start the proxy lifecycle management server
 	go m.startLifecycleServer()
+	go func() {
+		time.Sleep(time.Duration(m.startupGracePeriodSeconds) * time.Second)
+		m.gracefulTimeoutReached = true
```
sreeram77, line 112: This can lead to a race condition.

cmd/consul-dataplane/config.go
```diff
@@ -242,7 +242,7 @@ func buildDefaultConsulDPFlags() (DataplaneConfigFlags, error) {
 			"shutdownDrainListenersEnabled": false,
 			"shutdownGracePeriodSeconds": 0,
 			"gracefulShutdownPath": "/graceful_shutdown",
-			"gracefulPort": 20300,
+			"gracefulPort": 20600,
```
sreeram77, line 245: Why is this port changed?

pkg/consuldp/consul_dataplane.go
```diff
@@ -347,6 +348,7 @@ func (cdp *ConsulDataplane) GracefulShutdown(cancel context.CancelFunc) {
 	// If proxy lifecycle manager has not been initialized, cancel parent context and
 	// proceed to exit rather than attempting graceful shutdown
 	if cdp.lifecycleConfig != nil {
+		cdp.logger.Error("lifecycle config not initialized", "error")
```
sreeram77, line 351: logger requires args in pairs. We should be using (msg, argKey, argVal...).

cmd/consul-dataplane/main.go
```diff
@@ -166,6 +177,12 @@ func run() error {
 		return err
 	}
 
+	if subcommand == "graceful-startup" {
+		fmt.Println("graceful port is :", consuldpCfg.Envoy.GracefulPort)
```
sreeram77, line 181: Is Println required here if we're already logging it in the next line?

cmd/consul-dataplane/config_test.go
```diff
@@ -87,7 +87,7 @@ func TestConfigGeneration(t *testing.T) {
 						ShutdownDrainListenersEnabled: true,
 						GracefulShutdownPath:          "/graceful_shutdown",
 						EnvoyDrainTimeSeconds:         30,
-						GracefulPort:                  20300,
+						GracefulPort:                  20600,
```
anandmukul93, line 90: How did we get this new port id ?

cmd/consul-dataplane/graceful_startup.go
```diff
@@ -0,0 +1,33 @@
...
+	ticker := time.NewTicker(100 * time.Millisecond)
+	defer cancel()
+	defer ticker.Stop()
+
+	for {
+		select {
+		case <-ctx.Done():
+			return nil
+		case <-ticker.C:
+			resp, err := http.Get(fmt.Sprintf("http://localhost:%d%s", gracefulPort, gracefulStartupPath))
```
anandmukul93, line 23: can we extract literals to constants at top of file with meaningful names ?

cmd/consul-dataplane/graceful_startup.go
anandmukul93: I don't think fmt will print proper log with level and timestamp etc. I am not sure but do we configure fmt object somewhere else if we use it for logging even ?

cmd/consul-dataplane/main.go
```diff
@@ -166,6 +177,12 @@ func run() error {
...
+		log.Default().Printf(fmt.Sprintf("graceful port is: %d", consuldpCfg.Envoy.GracefulPort))
```
anandmukul93, line 182: can we use : 
github.com/hashicorp/go-hclog

example - https://github.com/hashicorp/consul/blob/5fb9df1640cfa1aaf96228da1f44b31ecd6b825a/agent/watch_handler.go#L99

pkg/consuldp/consul_dataplane.go
anandmukul93: > logger requires args in pairs. We should be using (msg, argKey, argVal...).
Suggestion : 
This was confusing in my case as well. We should have a strong typed wrapper map here for args. Can we create a small consul log wrapper library with multiple loggers support ? like logrus, zap, hclog, etc. Add runtime logging level change . Add support for stack trace as well.

PR #685, opened 2025-02-12
Upgraded envoy version in Dockerfile

nitin-sachdev-29:
Upgraded envoy version to 1.33.0 in Dockerfile


abhishek-hashicorp (approved):
LGTM

PR #701, opened 2025-03-13
update Go modules to fix CVEs

mrgupta-hashi:
Upgraded the following versions

go version to 1.23.0
golang.org/x/net v0.36.0
golang.org/x/sys v0.30.0
golang.org/x/text v0.22.0


abhishek-hashicorp (changes requested):
Please add the changelog. 
You should raise the PR to `main` and then add `backport/1.x` label to backport this to all the active release versions. This is a hacky way of doing it.

PR #702, opened 2025-03-13
update Go modules to fix CVEs

mrgupta-hashi:
Upgraded the following versions

golang.org/x/net v0.37.0
golang.org/x/sys v0.31.0
golang.org/x/text v0.23.0


nitin-sachdev-29 (approved):
LGTM

PR #759, opened 2025-06-23
upgrade go to 1.23.10

sujay-hashicorp:
## Changes
 
- Upgraded go version to 1.23.10
- Updated integration test matrix to run test with consul 1.21.x

# Tests

build and integration test run: https://github.com/hashicorp/consul-dataplane/actions/runs/15817095139


shore (dismissed):
build.yml changes look reasonable to me.  Nice work removing the version constraint. 😸

.github/workflows/build.yml
```diff
@@ -349,6 +349,8 @@ jobs:
             image: hashicorppreview/consul:1.19-dev
           - version: v1.20.0-dev
             image: hashicorppreview/consul:1.20-dev
+          - version: v1.21.0-dev
```
sreeram77, line 352: Unsupported consul versions (1.15) can be removed.

sujay-hashicorp: removed it now

sreeram77 (approved):

PR #1038, opened 2026-03-24
Upgrade debian11 to debian12 to address CVE

paras-gupta2:
## Description
- Upgrade debian11 to debian12 to address CVE's in prepare workflow


anilvpatel (approved):
LGTM

PR #1063, opened 2026-04-27
fix: add skip_exit to drain_listeners to prevent premature pod termination - 2

santoshpulluri:
### Description
When upgrading Envoy from 1.35 to 1.37, pods were being terminated early
during graceful shutdown. The POST /drain_listeners?graceful endpoint causes
Envoy to automatically exit after the drain period (--drain-time-s), which
raced against consul-dataplane's own lifecycle sequencing:

Drain() → wait shutdownGracePeriodSeconds → Quit() → Kill()

Adding &skip_exit decouples the drain phase from the exit decision, ensuring
Envoy stays alive until consul-dataplane explicitly calls Quit() via
/quitquitquit, preserving the intended shutdown lifecycle control.


panman90 (approved):
LGTM

PR #1066, opened 2026-04-28
fix: use net.JoinHostPort for remaining Envoy admin URLs

santoshpulluri:
### Description

Follow-up to [#1064](https://github.com/hashicorp/consul-dataplane/pull/1064).

This updates the remaining Envoy admin URLs in `pkg/envoy/proxy.go` to use `net.JoinHostPort` instead of manually formatting `host:port` pairs.

The previous `fmt.Sprintf("http://%s:%v/...")` pattern is not safe for all admin bind addresses, especially IPv6 addresses. `Drain()` and `Quit()` were already updated to use `net.JoinHostPort`; this change applies the same fix to:
- `DumpConfig()` via `/config_dump?include_eds`
- `Ready()` via `/ready`

This keeps Envoy admin endpoint construction consistent across the proxy manager and avoids malformed URLs when the admin address is not a simple IPv4/hostname value.

### Testing

- Ran `pkg/envoy/proxy_test.go`

### DIFF
--- a/.changelog/1066.txt
+++ b/.changelog/1066.txt
@@ -1,0 +1,2 @@
+```release-note:bug
+- envoy: use safe host:port construction for remaining admin API URLs to avoid malformed admin endpoint URLs with IPv6 addresses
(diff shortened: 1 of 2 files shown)


panman90 (approved):
LGTM

shashankNandigama (approved):
LGTM

PR #1142, opened 2026-06-02
fix(Dockerfile): fixing CVEs and best practice fixes 

santoshpulluri:
- Pin Alpine to 3.23 instead of unpinned 'latest'
- Add microdnf update and clean all to non-FIPS UBI stage (consistency with FIPS)
- Fix apt-get usage in setcap stages (use apt-get, --no-install-recommends, clean cache, combine layers)
- Add --no-cache to apk add in dumb-init stage
- Normalize all 'as' to 'AS' in FROM statements
- Add BIN_NAME default to FIPS stages for consistency


panman90 (dismissed):
LGTM

shashankNandigama (dismissed):
LGTM

Dockerfile
```diff
@@ -11,7 +11,7 @@
 # prebuilt binaries in any other form.
 #
 ARG GOLANG_VERSION
-FROM hashicorp/envoy:1.37.2 as envoy-binary
+FROM hashicorp/envoy:1.37.2 AS envoy-binary
```
shashankNandigama:
```suggestion
ARG ENVOY_VERSION=1.37.2
FROM hashicorp/envoy:${ENVOY_VERSION} as envoy-binary
```

similarly in line 31
FROM hashicorp/envoy-fips:${ENVOY_VERSION}-fips1402 as envoy-fips-binary

By adding this arg we can update the envoy version in a single place for fips and non-fips.

shashankNandigama: applied this change

vdinesh4738:
Santosh can you please add the suggestion from copilot

vdinesh4738 (approved):
LGTM

PR #1177, opened 2026-06-26
forwarding dns query to envoy

panman90:
Summary
This PR updates the DNS proxy to triage queries before forwarding them, so virtual Consul DNS queries can be resolved by Envoy when possible. It adds support for forwarding .virtual.*.consul queries to Envoy’s inline DNS listener, forwarding external domains to Envoy’s egress DNS listener, and falling back to Consul when Envoy cannot answer or is unavailable.

Changes
- introducing DNS query triage logic to classify requests as:
  virtual Consul domains
  non-virtual Consul domains
  external domains
- expand short-form virtual DNS names into canonical fully qualified Consul names before forwarding to Envoy
- rewrite successful Envoy responses back to the originally requested DNS name
- fall back to Consul DNS resolution when:
  Envoy returns NXDOMAIN for virtual queries
  Envoy listeners are unavailable
  request rewriting or forwarding fails
- add listener health backoff handling to avoid repeatedly forwarding to unhealthy Envoy listeners

## PCI review checklist

- [ ] I have documented a clear reason for, and description of, the change I am making.

- [ ] If applicable, I've documented a plan to revert these changes if they require more than reverting the pull request.

- [ ] If applicable, I've documented the impact of any changes to security controls.

  Examples of changes to security controls include using new access control methods, adding or removing logging pipelines, etc.

### DIFF
--- a/integration-tests/main_test.go
+++ b/integration-tests/main_test.go
@@ -274,6 +274,32 @@ func TestIntegration(t *testing.T) {
 		require.ElementsMatch(t, []string{backendPod.ContainerIP}, addrs)
 	}
 
+	// Virtual-domain (VIP) resolution: the frontend sidecar has "backend" as an
+	// upstream, so a *.virtual.consul query for it must be answered by Envoy's
+	// inline DNS listener with a Consul VIP from the 240.0.0.0/4 range
+	for _, port := range dnsPorts {
+		addrs, rcode := DNSLookupA(t,
+			suite,
+			port.Proto(),
+			frontendPod.HostIP,
+			frontendPod.MappedPorts[port],
+			"backend.virtual.consul.",
+		)
+
+		require.Equalf(t, DNSRcodeSuccess, rcode,
+			"expected backend.virtual.consul to resolve successfully over %s, got rcode=%d",
+			port.Proto(), rcode)
+		require.NotEmptyf(t, addrs,
+			"expected backend.virtual.consul to resolve to a VIP over %s, got no answers",
+			port.Proto())
+
+		for _, addr := range addrs {
+			require.Truef(t, IsConsulVIP(addr),
+				"expected backend.virtual.consul to resolve to a Consul VIP (240.0.0.0/4), got %q over %s",
+				addr, port.Proto())
+		}
+	}
+
 	metrics := GetMetrics(t,
 		backendPod.HostIP,
 		backendPod.MappedPorts[metricsPort],
--- a/integration-tests/main_test.go
+++ b/integration-tests/main_test.go
@@ -426,11 +426,19 @@ func TestIntegration(t *testing.T) {
 		backendPod.MappedPorts[upstreamLocalBindPort],
 	)
 
-	// BCDR: kill the Consul server to simulate a full control-plane outage and
-	// confirm that an already-configured *.virtual.consul domain continues to
-	// resolve via Envoy's inline DNS table with no connection to the control
-	// plane.
+	// virtual .virtual.consul resolves after consul server outage
+	//
+	// Consul servers >= v2.2 push Envoy DNS listeners (the inline virtual-DNS
+	// table) over xDS, so Envoy can continue answering *.virtual.consul queries
+	// from its local table after the control plane is gone. On older servers the
+	// listener is never pushed, so resolution after an outage is not expected to
+	// work. The test is therefore only asserted on v2.2+; on earlier versions it
+	// is skipped to avoid a spurious failure.
 	t.Run("virtual .virtual.consul resolves after consul server outage", func(t *testing.T) {
+		// Determine whether the running Consul server is new enough to push the
+		// inline virtual-DNS listener over xDS. The feature landed in v2.2.
+		supportsInlineDNS := semver.Compare(semver.MajorMinor(opts.ServerVersion), "v2.2") >= 0
+
 		serverContainerID := server.Container.GetContainerID()
 		dockerCli, err := client.NewClientWithOpts(client.WithAPIVersionNegotiation())
 		if err != nil {
@@ -441,35 +449,44 @@ func TestIntegration(t *testing.T) {
 		}
 		t.Logf("consul server container %s killed (SIGKILL) to simulate control-plane outage", serverContainerID)
 
+		if !supportsInlineDNS {
+			t.Skipf("skipping outage-resilience check: consul server %s < v2.2 does not push inline virtual-DNS listeners over xDS",
+				opts.ServerVersion)
+		}
+
 		// Query with retries: after a hard server kill the Envoy DNS proxy is
 		// still alive and serving the inline table, but the very first UDP
 		// exchange may time out while the proxy's upstream health-check to
 		// Consul drains. Retry for up to 15 s before failing.
-		for _, port := range dnsPorts {
-			addrs, rcode := DNSLookupA(t,
-				suite,
-				port.Proto(),
-				frontendPod.HostIP,
-				frontendPod.MappedPorts[port],
-				"backend.virtual.consul.",
-			)
+		require.Eventually(t, func() bool {
+			for _, port := range dnsPorts {
+				addrs, rcode, err := DNSLookupAErr(suite,
+					port.Proto(),
+					frontendPod.HostIP,
+					frontendPod.MappedPorts[port],
+					"backend.virtual.consul.",
+				)
+				if err != nil {
+					t.Logf("DNS virtual lookup transient error (retrying): consul_server_version=%s proto=%s err=%v",
+						opts.ServerVersion, port.Proto(), err)
+					return false
+				}
 
-			t.Logf("DNS virtual lookup: consul_server_version=%s proto=%s query=%s rcode=%d addrs=%v",
-				opts.ServerVersion, port.Proto(), "backend.virtual.consul.", rcode, addrs)
+				t.Logf("DNS virtual lookup: consul_server_version=%s proto=%s query=%s rcode=%d addrs=%v",
+					opts.ServerVersion, port.Proto(), "backend.virtual.consul.", rcode, addrs)
 
-			require.Equalf(t, DNSRcodeSuccess, rcode,
-				"expected backend.virtual.consul to resolve successfully over %s, got rcode=%d",
-				port.Proto(), rcode)
-			require.NotEmptyf(t, addrs,
-				"expected backend.virtual.consul to resolve to a VIP over %s, got no answers",
-				port.Proto())
-
-			for _, addr := range addrs {
-				require.Truef(t, IsConsulVIP(addr),
-					"expected backend.virtual.consul to resolve to a Consul VIP (240.0.0.0/4), got %q over %s",
-					addr, port.Proto())
+				if rcode != DNSRcodeSuccess || len(addrs) == 0 {
+					return false
+				}
+				for _, addr := range addrs {
+					if !IsConsulVIP(addr) {
+						return false
+					}
+				}
 			}
-		}
+			return true
+		}, 15*time.Second, 1*time.Second,
+			"expected backend.virtual.consul to resolve to a Consul VIP via Envoy inline DNS table after control-plane outage")
 	})
 
 	// Send SIGTERM to dataplane to start graceful shutdown
--- a/integration-tests/main_test.go
+++ b/integration-tests/main_test.go
@@ -336,48 +336,6 @@ func TestIntegration(t *testing.T) {
 		}
 	})
 
-	// BCDR: kill the Consul server to simulate a control-plane outage and
-	// confirm that an already-configured *.virtual.consul domain continues to
-	// resolve via Envoy's inline DNS table without any connection to the control
-	// plane.
-	t.Run("virtual .virtual.consul resolves after consul server outage", func(t *testing.T) {
-		serverContainerID := server.Container.GetContainerID()
-		dockerCli, err := client.NewClientWithOpts(client.WithAPIVersionNegotiation())
-		if err != nil {
-			t.Fatalf("failed to create docker client: %v", err)
-		}
-		if err := dockerCli.ContainerKill(context.Background(), serverContainerID, "SIGKILL"); err != nil {
-			t.Fatalf("failed to kill consul server container %s: %v", serverContainerID, err)
-		}
-		t.Logf("consul server container %s killed (SIGKILL) to simulate control-plane outage", serverContainerID)
-
-		for _, port := range dnsPorts {
-			addrs, rcode := DNSLookupA(t,
-				suite,
-				port.Proto(),
-				frontendPod.HostIP,
-				frontendPod.MappedPorts[port],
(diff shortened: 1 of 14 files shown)


pkg/dns/upstream_index.go
```diff
@@ -0,0 +1,174 @@
+// Copyright IBM Corp. 2022, 2026
```
panman90, line 1: This changes required to read partition, datacenter, NS information for upstream api from CDS frame as while FQDN expansion data-plane is not aware of this information of upstreams

pkg/dns/dns.go
```diff
@@ -345,3 +334,306 @@ func (d *DNSServer) Stop() {
...
+	switch class {
+	case domainClassConsul:
+		// Standard Consul domain — unchanged path.
+		return d.queryConsul(raw, proto)
+
+	case domainClassExternal:
+		// Non-consul domain.
+		if !d.canTryEgressListener() {
+			return d.queryConsul(raw, proto)
+		}
```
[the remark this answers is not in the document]
panman90: taken care

pkg/dns/dns.go
```diff
@@ -345,3 +334,306 @@ func (d *DNSServer) Stop() {
...
+			d.markEgressListenerUnavailable()
+			d.logger.Debug("egress listener unavailable, falling back to consul", "error", err)
+			return d.queryConsul(raw, proto)
+		}
+		return resp, nil
+
+	case domainClassVirtual:
+		// Expand the short form to the full FQDN.
+		expandedName := expandVirtualFQDN(originalName, d.namespace, d.partition, d.datacenter)
+
```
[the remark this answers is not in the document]
panman90: taken care

pkg/dns/dns.go
```diff
@@ -290,45 +295,29 @@ func (d *DNSServer) proxyTCPAcceptedConn(ctx context.Context, conn net.Conn, cli
...
-			"x-consul-namespace", d.namespace,
-			"x-consul-token", d.token,
-		)
-
-		logger.Debug("querying through tcp", "partition", d.partition, "namespace", d.namespace)
-
-		resp, err := client.Query(ctx, req)
+		logger.Debug("triaging dns request", "partition", d.partition, "namespace", d.namespace)
+		responseMsg, err := d.triageAndResolve(data, pbdns.Protocol_PROTOCOL_TCP)
 		if err != nil {
```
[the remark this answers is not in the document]
panman90: taken care

pkg/dns/dns.go
bharath-k1999, line 256: When `listenerTCP.Accept()` returns an error, `c` is `nil`. The code logs the warning but still launches `go proxyTCPAcceptedConn(ctx, c)`. The goroutine's first line is `defer conn.Close()` — this panics with a nil-pointer dereference on any TCP accept error (e.g. transient resource exhaustion).

```go
// Current — nil conn dispatched on error
c, err := d.listenerTCP.Accept()
if err != nil {
    d.logger.Warn("failure to accept tcp connection", "error", err)
}
go d.proxyTCPAcceptedConn(ctx, c) // ← panics if c == nil

// Fix — mirror proxyUDP's pattern (line 217)
c, err := d.listenerTCP.Accept()
if err != nil {
    if errors.Is(err, net.ErrClosed) {
        return // listener closed, exit loop cleanly
    }
    d.logger.Warn("failure to accept tcp connection", "error", err)
    continue // do not dispatch with a nil conn
}
go d.proxyTCPAcceptedConn(ctx, c)

File: integration-tests/main_test.go [hunk 5]
```diff
@@ -263,16 +263,104 @@ func TestIntegration(t *testing.T) {
...
-		addrs := DNSLookup(t,
-			suite,
-			port.Proto(),
-			frontendPod.HostIP,
-			frontendPod.MappedPorts[port],
-			"backend-sidecar.service.consul.",
-		)
-		require.ElementsMatch(t, []string{backendPod.ContainerIP}, addrs)
-	}
+	adminIP := frontendPod.HostIP
```
bharath-k1999, line 266: RFC reference: Testing Strategy: "End-to-end / BCDR: Simulate control-plane outage and confirm the data plane continues resolving configured virtual domains with existing config."

The integration suite verifies *.virtual.consul resolves to a VIP when both the dataplane and Consul server are healthy. But there is no test that stops the Consul control plane and then asserts that an already-configured virtual domain still resolves via the Envoy inline table.

This is the headline BCDR guarantee of the entire RFC. Without it, a future refactor (e.g. dropping UpstreamIndex state on reconnect) could silently break the guarantee undetected.

The harness already uses client.ContainerKill for the graceful-shutdown test (line 438) — the same pattern can stop the Consul server container and re-run the DNSLookupA + IsConsulVIP assertion on the virtual domain.

pkg/dns/dns_test.go
```diff
@@ -330,3 +331,348 @@ func (s *DNSTestSuite) Test_ProxydnsTCP() {
...
+			Once()
+
+		resp, err := server.triageAndResolve(query, pbdns.Protocol_PROTOCOL_UDP)
+		s.Require().NoError(err)
+		s.Require().Equal(consulResp, resp)
+		s.Require().False(server.canTryEgressListener())
+	})
+}
+
+func (s *DNSTestSuite) Test_TriageAndResolve_VirtualDomain() {
```
bharath-k1999, line 440: The RFC explicitly names this as a design challenge and mandates that the dataplane handles all 8 alias forms. The unit tests cover only two cases in Test_TriageAndResolve_VirtualDomain:

service.virtual.consul (bare, no tokens)
service-db.service.virtual.consul (with .service. alias)
The following 6 combinations produced by parseVirtualTokens have no direct test:

svc.virtual.<ns>.ns.consul — namespace only
svc.virtual.<ap>.ap.consul — partition only
svc.virtual.<dc>.dc.consul — datacenter only
svc.virtual.<ns>.ns.<ap>.ap.consul — ns + partition, no dc
svc.virtual.<ns>.ns.<dc>.dc.consul — ns + dc, no partition
svc.virtual.<ap>.ap.<dc>.dc.consul — partition + dc, no ns
A table-driven test for parseVirtualTokens and expandVirtualName (similar to TestParseServiceSNI in upstream_index_test.go) would directly satisfy the RFC's unit-testing requirement for FQDN expansion and prevent regressions in the token-parsing logic.

pkg/dns/upstream_index.go
```diff
@@ -0,0 +1,174 @@
...
+//	other partition, with subset:   <subset>.<svc>.<ns>.<ap>.<dc>.internal-v1.<trustdomain>
+//
+// The subset label, when present, is ignored: virtual-FQDN expansion keys off
+// the base service name, and every subset of a service shares its identity.
+//
+// This parsing is intentionally coupled to Consul's SNI label scheme (see
+// agent/connect/sni.go in the Consul server). SNIs that are external, peered,
+// gateway, or prepared-query are not internal upstreams and return ok=false.
+func ParseServiceSNI(sni string) (UpstreamComponents, bool) {
+	sni = strings.ToLower(strings.TrimSuffix(sni, "."))
```
sanikachavan5, line 193: are we not considering multiport services?

sanikachavan5, line 193: even, https://github.com/hashicorp/consul/blob/main/agent/xds/clusters.go#L427 passthroughs, directly splitting and taking the first position can be erroneous in some specific cases.

sanikachavan5 (changes requested):

pkg/dns/upstream_index.go
```diff
@@ -0,0 +1,199 @@
...
+	if sni == "" {
+		return UpstreamComponents{}, false
+	}
+
+	// Strip any cluster-name prefix that precedes the bare SNI.
+	// These prefixes are separated from the SNI by "~", not ".", so they never
+	// appear as a dot-split label. A single pass is enough; prefixes do not nest.
+	for _, pfx := range sniClusterPrefixes {
```
sanikachavan5: Could we also handle Consul’s customization-hash prefix here? naming.CustomizeClusterName prefixes CDS resource names as <hash>~<cluster-name>, for example f8f8f8f8~pong.default.dc2.internal.... For a single-port cluster without a subset, the parser below consequently extracts f8f8f8f8~pong as the service name, so Lookup("pong", ...) misses and short virtual names may be expanded using the source proxy’s defaults instead of the canonical inline-table identity.
Multiport names happen to work because the decorated port label is ignored when parsing backward from internal. Could we strip the customization hash here—including after passthrough~ or exported~—and add tests for the single-port and combined-prefix cases?
Consul generates this prefix here: https://github.com/hashicorp/consul/blob/59ad54c2770fac4fe8cb9b23bbf29175752d1675/agent/xds/naming/naming.go#L20-L27

sanikachavan5 (changes requested):

pkg/dns/dns.go
```diff
@@ -345,3 +346,377 @@ func (d *DNSServer) Stop() {
...
+	return n
+}
+
+// isNXDOMAIN returns true when the DNS response carries an NXDOMAIN rcode.
+func isNXDOMAIN(raw []byte) bool {
+	var msg dnsmessage.Message
+	if err := msg.Unpack(raw); err != nil {
+		return false
+	}
+	return msg.RCode == dnsmessage.RCodeNameError
```
sanikachavan5: What happens if the err is not a RcodeNameError, but is a malformed request, we will still process it through.

sanikachavan5 (changes requested):

pkg/consuldp/xds.go
```diff
@@ -118,15 +120,23 @@ func (cdp *ConsulDataplane) xdsServerExited() chan struct{} { return cdp.xdsServ
...
+	// upstreamIndex, when set, receives CDS cluster SNIs decoded from the
+	// server->Envoy frames that pass through this stream.
+	upstreamIndex *dns.UpstreamIndex
 }
 
 func (s *metricServerStream) SendMsg(m interface{}) error {
+	// Tap CDS frames on their way to Envoy to keep the upstream identity index
+	// current. This reads a copy of the frame's bytes and never mutates m, so
+	// opaque passthrough is unaffected.
+	tapClusterFrame(s.upstreamIndex, m)
```
sanikachavan5: Could we update the upstream index only after ServerStream.SendMsg succeeds? Currently a failed send still applies the CDS delta locally, so dataplane may expand short DNS names using an identity that Envoy never received while Envoy retains its previous inline DNS table. That mismatch can turn previously resolvable names into NXDOMAIN during an xDS/control-plane interruption. Updating after a successful send would at least prevent indexing frames that were definitely not delivered.

pkg/dns/dns.go
```diff
@@ -345,3 +346,377 @@ func (d *DNSServer) Stop() {
...
+	domainClassExternal                    // everything else
+)
+
+// classifyDomain returns the class of the provided DNS query name.
+// The name is normalized to lower-case and any trailing '.' is removed.
+func classifyDomain(name string) domainClass {
+	name = strings.ToLower(strings.TrimSuffix(name, "."))
+	if strings.HasSuffix(name, ".consul") || name == "consul" {
+		// Check for virtual: must contain ".virtual." segment
+		if strings.Contains(name, ".virtual.") {
```
sanikachavan5: Could we validate the complete Consul DNS grammar here instead of treating every .consul name containing .virtual. as a virtual lookup?
I reproduced a collision using an image built from the exact companion Consul PR head, 59ad54c2770fac4fe8cb9b23bbf29175752d1675. 

I registered this Connect-native mesh service:
`{
  "Name": "virtual",
  "ID": "virtual-mesh-1",
  "Address": "127.0.0.42",
  "Port": 8080,
  "Tags": ["blue"],
  "Connect": {"Native": true}
}`

Consul accepted the registration and reported Connect.Native: true. The standard tagged-service query:
`dig @127.0.0.1 -p 18600 blue.virtual.service.consul A`
returned:
blue.virtual.service.consul. 0 IN A 127.0.0.42
Consul interprets this as tag blue, service virtual, because service is the rightmost query-kind label. However, adding this dataplane test:
"blue.virtual.service.consul": domainClassConsul,
fails because classifyDomain returns domainClassVirtual.

This check runs before mesh cluster, named-port, or ALPN/SNI selection, so the same misclassification also applies to multiport mesh services: later multiport handling cannot correct a query that has already been routed down the virtual-DNS path. If blue is an inline upstream, this query could return blue’s VIP instead of resolving the service named virtual.

Could we make parseVirtualTokens strictly consume one of the supported virtual-domain forms and use that result for classification? It currently also accepts any .virtual. occurrence and ignores unrecognized remaining labels. Please add this collision as a regression test.

sanikachavan5 (changes requested):

pkg/dns/dns.go
```diff
@@ -345,3 +346,377 @@ func (d *DNSServer) Stop() {
...
+// fill them). ok is false when name is not a virtual domain.
+func parseVirtualTokens(name string) (svc, ns, partition, dc string, ok bool) {
+	name = strings.ToLower(strings.TrimSuffix(name, "."))
+
+	// Find the service name: everything up to the first ".virtual." segment.
+	virtualIdx := strings.Index(name, ".virtual.")
+	if virtualIdx < 0 {
+		return "", "", "", "", false
+	}
+	svc = name[:virtualIdx]
```
sanikachavan5: Could we handle Consul’s named-port virtual DNS form here? Consul supports <port>.<service>.virtual.consul; for example, http.api.virtual.consul is parsed as port http and service api ([Consul DNS implementation](https://github.com/hashicorp/consul/blob/59ad54c2770fac4fe8cb9b23bbf29175752d1675/agent/dns.go#L911-L929)).

I validated the current dataplane parser with:
svc, _, _, _, ok := parseVirtualTokens("http.api.virtual.consul")
require.True(t, ok)
require.Equal(t, "api", svc)
The test fails because the actual service is "http.api". parseVirtualTokens treats everything before the first .virtual. as the service name.

The companion Consul PR builds the inline-table key from the base uid.Name, producing api.virtual.<ns>.ns.<partition>.ap.<dc>.dc.consul, not http.api.virtual... ([inline FQDN generation](https://github.com/hashicorp/consul/blob/59ad54c2770fac4fe8cb9b23bbf29175752d1675/agent/xds/listeners_dns.go#L282-L309)). Consequently, a port-qualified lookup expands to a name that cannot exist in Envoy’s inline table and falls back to the control plane; during a control-plane outage it fails completely.
This gap applies whenever the named-port virtual form is used, whether the multiport service comes from catalog service discovery or is consumed as a mesh upstream.
 The base service-level form api.virtual.consul may still work, but existing port-qualified Consul DNS compatibility does not.

Could we split the optional port label from the base service name and expand the inline lookup using only api, since the RFC assigns one VIP per service? Please add regression coverage for <port>.<service>.virtual.consul and its namespace/partition/datacenter variants.

sanikachavan5 (changes requested):
the entire code piece

sanikachavan5 (approved):
looks good to me, but integration tests seem to fail, please take a look at that.

PR #1198, opened 2026-07-21
Removal of noisy log

shani34:
consul-dataplane.dns-proxy.udp: timeout waiting for read: error="read udp nnn i/o timeout"

### DIFF
--- a/pkg/dns/dns.go
+++ b/pkg/dns/dns.go
@@ -189,7 +189,7 @@ func (d *DNSServer) proxyUDP(ctx context.Context) {
 			if errors.Is(err, net.ErrClosed) {
 				logger.Info("connection closed")
 				return
-			} else if nerr, ok := err.(net.Error); !ok && !nerr.Timeout() {
+			} else if nerr, ok := err.(net.Error); !ok || !nerr.Timeout() {
 				logger.Warn("error reading from conn", "error", err)
 			}
 			continue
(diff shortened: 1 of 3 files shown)


pkg/dns/dns.go
```diff
@@ -189,9 +189,7 @@ func (d *DNSServer) proxyUDP(ctx context.Context) {
 			if errors.Is(err, net.ErrClosed) {
 				logger.Info("connection closed")
 				return
-			} else if nerr, ok := err.(net.Error); ok && nerr.Timeout() {
-				logger.Debug("timeout waiting for read", "error", err)
-			} else {
+			} else if nerr, ok := err.(net.Error); !ok && !nerr.Timeout() {
```
rashmirathi-2: Please add test cases.
Should this be && or ||

pkg/dns/dns_test.go
```diff
@@ -225,6 +266,39 @@ func (s *DNSTestSuite) Test_UDPProxy() {
 
 }
 
+func (s *DNSTestSuite) Test_UDPProxy_ReadErrorContinues() {
```
sanikachavan5: The test misses the behavior being changed. In both cases are non-timeout errors and the test only verifies loop continuation. It would pass if timeout logging were accidentally restored. Add a net.Error with Timeout() == true and capture the logger to assert that neither "timeout waiting for read" nor the warning is emitted.

sanikachavan5 (changes requested):

pkg/dns/dns_test.go
shani34: updated test case, please check

sanikachavan5 (approved):

PR #1293, opened 2026-09-11
updating dns failure

panman90:
Reworks DNS listener failure backoff tracking to use atomic timestamps instead of mutex-protected time.Time values.

Changes:

Replaces listener health mutex with atomic deadline fields.
Updates listener availability checks and failure marking.


nitin-sachdev-29:
@panman90  Please add a proper description.

nitin-sachdev-29 (approved):
LGTM

PR #1294, opened 2026-09-15
feat(oauth): add local credential broker for split-knowledge DEKs

chittaranjanswainibm:
## Summary

Adds an optional **Local Credential Broker** inside `consul-dataplane` so CAMP AI-agent workloads can obtain DEKs for OBO (`private_key_jwt`) without putting plaintext signing material on xDS/SDS.

When `-credential-broker-bind-addr` is set (typically a pod-local Unix socket), dataplane:

1. Binds **Unix-only** (TCP/other schemes rejected); listen failure **aborts startup**
2. Serves `LocalCredentialBroker.GetKey` over that UDS
3. On cache miss/refresh, calls Consul `WorkloadKeyService.FetchKey` on the existing dataplane gRPC session (proxy identity + ACL — not a caller-supplied workload id)
4. Caches DEKs in memory; enforces same-UID peer creds on Linux/Darwin
5. On unexpected broker serve exit, tears down the dataplane (same fail-closed pattern as xDS)

Default (empty bind addr) leaves OSS dataplane behavior unchanged.

### `pkg/credentialbroker/`
- New in-memory broker: `GetKey` / `GetBrokerStatus`, refresh fraction, expiry handling
- Per-key fetch coalescing; stale-cache fallback when refresh returns an expired key
- Nil-safe `Fetcher` / `FetchKeyResponse`; prune expired cache entries on `put`
- UDS listen with `0660`, peer-cred UID check (`SO_PEERCRED` / `LOCAL_PEERCRED`), stub for other OS
- `Exited()` channel for unexpected serve failure
- Unit tests for cache, expiry, coalescing, peer path rejection, nil inputs

### `pkg/consuldp/`
- `CredentialBrokerConfig`; `startCredentialBroker` wires `WorkloadKeyServiceClient` + proxy identity into `FetchKey`
- `Run` starts the broker after bootstrap params and selects on `credentialBrokerExited()`

### `cmd/consul-dataplane/`
- Flags/env: `-credential-broker-bind-addr`, `-credential-broker-refresh-fraction` (default `0.2`)
- Config merge / `Float64Var` helpers + tests

### `internal/pbdataplane/`
- Protos + generated stubs: `LocalCredentialBroker`, `WorkloadKeyService` (client-side copy so dataplane does not wait on a Consul release)

### Other
- `.changelog/1294.txt`
- Coverage workflow excludes `*.pb.go` from PR coverage denominator

## Breaking Changes & Backward Compatibility

**Not a breaking change.**

- Broker is **opt-in**; empty bind addr → no socket, no new RPCs, no behavior change
- No changes to existing xDS/Envoy/DNS public interfaces
- New flags/env vars are additive
- Peer-cred enforcement only applies when the broker is enabled

## How Has This Been Tested?

**Unit tests**
- `go test ./pkg/credentialbroker ./pkg/consuldp ./cmd/consul-dataplane`
- Coverage includes: round-trip `GetKey` over UDS, expired → `FailedPrecondition`, missing key → `NotFound`, TCP bind rejected, listen parent-as-file fails, refresh coalescing, expired-refresh keeps valid cache, nil fetcher/response, peercred accept/reject, flag/env wiring, fetcher copies proxy identity

**Manual / integration (recommended)**
1. Flag unset → broker does not start; dataplane starts as today
2. Bind `unix:///…/credential-broker.sock` → socket appears; `GetKey` succeeds for a known `key_id`
3. Listen failure (bad path) → dataplane startup fails
4. Against consul-enterprise with `WorkloadKeyService`: FetchKey succeeds for this proxy identity; denied/unknown → `NotFound` (no ACL leak)
5. Kill/break the UDS server mid-run → dataplane exits (fail-closed)

## Cross-Repo Dependencies

| Repo | Dependency | Status / notes |
|------|------------|----------------|
| **consul-enterprise** | `#13453` — server `WorkloadKeyService.FetchKey`, envelope xDS, DCR gated by `EnableServiceDCR` + `ai.role` | Required for end-to-end FetchKey; this PR ships client protos so dataplane can merge first |
| **consul-dataplane** | **This PR `#1294`** | Local Credential Broker |
| **consul-k8s** | Inject `-credential-broker-bind-addr` / OBO `--broker-uds` for `consul.hashicorp.com/ai-role: ai-agent` | Follow-up; without it the broker stays off |
| **consul-ai-apps** | `consul-obo-outbound` envelope watcher + `GetKey` client | Follow-up consumer of this UDS |

`GetBrokerStatus` is implemented for optional health; OBO is expected to call `GetKey` only.


pkg/credentialbroker/broker.go
```diff
@@ -0,0 +1,357 @@
...
+	if rec := b.get(keyID); rec != nil {
+		if isExpired(rec, now) {
+			b.drop(keyID)
+			return nil, errKeyExpired
+		}
+		if !needsRefresh(rec, now) {
+			return toGetKey(rec.resp), nil
+		}
+	}
+	fetched, err := b.cfg.Fetcher.FetchKey(ctx, keyID)
```
nitin-sachdev-29: **[P2] The broker can return a key that expired while being fetched**

`doFetch` [records the current time before requesting the key](https://github.com/hashicorp/consul-dataplane/blob/2b8434dac8ccf78b2b923f74c62fef467776e1ba/pkg/credentialbroker/broker.go#L240-L251). After the response arrives, it checks expiry against that earlier time rather than the time of return.

A key that expires while the request is in flight can therefore be cached and returned as valid. The RFC requires expired keys to be rejected.

**Suggested fix:** Check expiry using the current time after the fetch completes and before returning the key.

nitin-sachdev-29 (commented):
Review findings from the CAMP RFC/HLD review.

nitin-sachdev-29 (changes requested):

nitin-sachdev-29 (approved):
LGTM

